PR #11, opened 2024-01-16
Fix issue with negative epoch

sosna:
There is a bug affecting the conversion of negative UNIX timestamps (i.e. dates before 1970-01-01) when deserializing fusion-json payloads.

The purpose of this PR is to fix this issue.


javihern98 (approved):
Code is OK, did not know FMR supported negative epoch.

stratosn (approved):

PR #18, opened 2024-02-22
Flatten hierarchy

sosna:
A utility method has been added, returning all the codes from a hierarchy as a flat list without duplicated entries.

This is useful in case the hierarchy is associated to a component in a dataflow and must be used for validation purposes.


javihern98 (approved):
Nice improvement on codes!

PR #20, opened 2024-02-29
Fix get_schema() query for provisionagreement context.

bvitez:


stratosn (commented):
I think that this is not enough, because even like that, it will fail if a Hierarchy Association is not found for the specific Dataflow, because the fetcher will raise an exception.
Unless the Hierarchy Association is mandatory (which I think is not) we need to handle the lack of HAs for any Dataflow (or Provision Agreement).

stratosn (approved):
Approving to unblock the PA case and will follow-up to fix the issue, but also to handle lack of HAs for Dataflows.

bvitez:
Sounds good. Thank you.

Issue #25, opened 2024-03-19
Model changes to class hierarchy


javihern98:
Generating code proposal using class hierarchy to adapt it to the SDMX-IM (as is pure form). Modify convenience methods and classes to adopt this structure.

sosna:
I'm closing this issue, as it was implemented and merged yesterday. Thanks for the nice PR and code, @javihern98

PR #27, opened 2024-03-19
SDMX IM base classes

javihern98:
Discussion over #25


src/pysdmx/model/code.py
```diff
@@ -75,30 +77,9 @@ class Codelist(Struct, frozen=True, omit_defaults=True):
         codes: The list of codes in the codelist.
     """
 
-    id: str
-    name: str
-    agency: str
-    description: Optional[str] = None
-    version: str = "1.0"
-    codes: Sequence[Code] = ()
+    items: Sequence[Code] = ()
```
stratosn, line 77: We need to align the documentation "codes" with the implementation "items".
As discussed, my preference is to keep the "codes", but I am happy otherwise.

javihern98: Yes, I am still struggling a lot with this specific change, as many tests fail and need to modify the whole process, I will probably try to rollback this change and use just codes in it, as this was just to align with the IM.

src/pysdmx/model/concept.py
```diff
@@ -181,28 +179,7 @@ class ConceptScheme(Struct, frozen=True, omit_defaults=True):
...
-        return len(self.concepts)
-
-    def __getitem__(self, id_: str) -> Optional[Concept]:
-        """Return the concept identified by the given ID."""
-        out = list(filter(lambda concept: concept.id == id_, self.concepts))
-        if len(out) == 0:
-            return None
-        else:
-            return out[0]
+    items: Sequence[Concept] = ()
```
stratosn: As in Codelists, my preference would be to use concepts.

javihern98: Yes, please check latest commit that solves this, this is outdated code

javihern98:
@sosna @stratosn FYI The parts that may break compatibility with previous versions are the changes between codes/concepts/categories and items (ItemScheme objects need to be initialised with items instead of codes/concepts/categories). As we agreed yesterday the codes/concepts/categories in ItemScheme should be a getter only, therefore they are not included in the class initialization. Please review this. I have already changed the internal serializers and a couple of tests to ensure everything works fine.

src/pysdmx/model/__base.py
```diff
@@ -0,0 +1,212 @@
+from datetime import datetime
+from typing import Any, Dict, Optional, Sequence, Union
+
+from msgspec import Struct
+
+
+class Annotation(Struct, frozen=True, omit_defaults=True):
```
sosna, line 7: We could add a post init check, making sure that at least one of the properties has been populated? Else, it would be possible to create totally empty annotations. Just a minor comment so, feel free to ignore 😊

src/pysdmx/model/__base.py
```diff
@@ -0,0 +1,212 @@
...
+    Provides identity to all derived classes. It also provides annotations
+    to derive classes because it is a subclass of AnnotableArtefact.
+
+    Attributes:
+        id: The identifier of the artefact.
+        uri: The URI of the artefact.
+        urn: The URN of the artefact.
+    """
+
+    id: str = ""
```
sosna, line 82: Why is the id defaulting to an empty string?

src/pysdmx/model/__base.py
```diff
@@ -0,0 +1,212 @@
...
+        service_url: The URL of the service.
+        structure_url: The URL of the structure.
+        agency: The maintainer of the artefact.
+    """
+
+    is_final: bool = False
+    is_external_reference: bool = False
+    service_url: Optional[str] = None
+    structure_url: Optional[str] = None
+    agency: Optional[Union[str, "Agency"]] = None
```
sosna, line 137: I think agency should not be optional for maintainable artefacts?

src/pysdmx/model/__base.py
```diff
@@ -0,0 +1,212 @@
...
+    """Organisation class.
+
+    Attributes:
+        contacts: The contact of the agency.
+    """
+
+    contacts: Sequence["Contact"] = ()
+
+
+class Agency(Organisation, frozen=True, omit_defaults=True):
```
sosna, line 175: We could also add DataProvider and DataConsumer for good measure. Happy to do it, once this PR has been merged :)

src/pysdmx/model/__base.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class Organisation(Item, frozen=True, omit_defaults=True):
+    """Organisation class.
+
+    Attributes:
+        contacts: The contact of the agency.
+    """
+
+    contacts: Sequence["Contact"] = ()
```
sosna, line 172: If we move Contact above, I guess we don't need to use the quotes here?

src/pysdmx/model/__base.py
```diff
@@ -0,0 +1,212 @@
...
+
+    The descriptive information for an arrangement or division of objects
+    into groups based on characteristics, which the objects have in common.
+
+    Attributes:
+        items: The list of items in the scheme.
+        is_partial: Whether the scheme is partial.
+    """
+
+    items: Sequence["Item"] = ()
```
sosna, line 161: Maybe we don't need to use quotes around Item, as it is available above?

src/pysdmx/model/__base.py
sosna: If we move Contact, Item, Organisation, Agency, etc above, I guess we don't need to use the quotes here?

src/pysdmx/model/category.py
```diff
@@ -6,12 +6,11 @@
...
 from typing import Iterator, List, Optional, Sequence, Set
 
-from msgspec import Struct
-
+from pysdmx.model import Item, ItemScheme
 from pysdmx.model.organisation import DataflowRef
 
 
-class Category(Struct, omit_defaults=True):
+class Category(Item, frozen=False):  # type: ignore[misc]
```
sosna, line 13: Why has omit_defaults been removed?

src/pysdmx/model/category.py
```diff
@@ -75,15 +64,12 @@ class CategoryScheme(Struct, frozen=True, omit_defaults=True):
...
 
-    id: str
-    name: str
-    agency: str
-    description: Optional[str] = None
-    version: str = "1.0"
-    categories: Sequence[Category] = ()
+    @property
+    def categories(self) -> Sequence[Category]:
+        """Extract the items in the Category Scheme."""
```
sosna, line 71: I would use the old documentation ("The list of top level categories in the scheme."), as it made it explicit that it would be the top level categories only (i.e. no recursive traversal of the categories and their descendants)

sosna (approved):
Hi @javihern98. I have reviewed the PR and made a few comments (all minor). It looks very good and we are going in the right direction, I believe 😊

src/pysdmx/model/__base.py
javihern98: This is to prevent mypy errors:

image

We could add a misc ignore to prevent this if needed. Defaulting id to an empty string may raise some issues for the final user, so it´s better to enforce the use of it. Do you agree?

javihern98: Need to add also kw_only=True to AnnotableArtefact to prevent misuse of the arguments to init all classes, it is also inherited by the subclasses.

src/pysdmx/model/__base.py
javihern98: Yes, please do it :)

src/pysdmx/model/category.py
javihern98: Adding it to the Docstring of this method.

src/pysdmx/model/__base.py
javihern98: Agree on Agency should not be optional, what should it be the default value for Agency? (None?, ""?). 

After some reorganisation I moved them to prevent the use of quotes in this file.

src/pysdmx/model/__base.py
javihern98: After some reorganisation I moved them to prevent the use of quotes in this file.

src/pysdmx/model/__base.py
javihern98: After some reorganisation I moved them to prevent the use of quotes in this file.

src/pysdmx/model/__base.py
javihern98: @sosna please add this after the PR.

src/pysdmx/model/category.py
javihern98: Solved on new commit. Just forgot when changing the class heirarchy.

src/pysdmx/model/__base.py
javihern98: Removed quotes on https://github.com/bis-med-it/pysdmx/pull/27/commits/40e23ff46b381c0287356ee437eac75fe92eedb0

src/pysdmx/model/__base.py
javihern98: Mentioned in #32

src/pysdmx/model/__base.py
javihern98: Mentioned in #33

src/pysdmx/model/__base.py
sosna: > We could add a misc ignore to prevent this if needed. Defaulting id to an empty string may raise some issues for the final user, so it´s better to enforce the use of it. Do you agree? @sosna
Yes, indeed, I agree. We should add a post init check to make sure the id has not been left empty

src/pysdmx/model/__base.py
javihern98: Okay, let´s go for it

javihern98: src\pysdmx\model\__base.py:197:34: error: Incompatible types in assignment (expression has type "None", variable has type "str | Agency")  [assignment]

Cannot have None as default value if we remove the Optional typing. Should we ignore it? @sosna

sosna: > Cannot have None as default value if we remove the Optional typing. Should we ignore it? @sosna 

You are right. Maybe we should default to an empty string then, and, as discussed, check in the post init function that a value has been set?

javihern98: Yes, adding a new issue with this.

javihern98: Mentioned in #34

javihern98:
https://github.com/bis-med-it/pysdmx/pull/27/commits/74a3d56bfe2deb4e5654560bf38ab42cf51abf48 resolved all conversations and added new issues #32, #33, #34 after discussing.

PR #36, opened 2024-05-03
Add post init validation checks

### LINKED ISSUE #32: Add post init check to Annotation
@sosna please add this after the PR.

_Originally posted by @javihern98 in https://github.com/bis-med-it/pysdmx/pull/27#discussion_r1587325659_

Adding a post init check to ensure an annotation is not generated without an ID.

### LINKED ISSUE #34: Add post init check to Agency in MaintainableArtefact
Check that Agency is set as a post init check.

sosna:
Post init validation checks have been added to:
- Annotations, to make sure at least one property is set
- Maintainable artefacts, to make sure the agency is set

Close #32
Close #34

### DIFF
--- a/src/pysdmx/model/__base.py
+++ b/src/pysdmx/model/__base.py
@@ -29,7 +29,7 @@ class Annotation(Struct, frozen=True, omit_defaults=True):
     url: Optional[str] = None
     type: Optional[str] = None
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
         """Additional validation checks for Annotation."""
         if (
             not self.id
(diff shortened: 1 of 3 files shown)


src/pysdmx/model/__base.py
```diff
@@ -27,6 +29,24 @@ class Annotation(Struct, frozen=True, omit_defaults=True):
     url: Optional[str] = None
     type: Optional[str] = None
 
+    def __post_init__(self) -> None:
+        """Additional validation checks for Annotation."""
+        if (
+            not self.id
```
javihern98, line 35: Maybe some general method is needed to avoid this combination, I will try to look for sth but is not blocking.

javihern98 (approved):
Just need a minor improvement on a general method to check if the class is completely empty.

src/pysdmx/model/__base.py
sosna: Thanks, @javihern98. My understanding is that you will search for another way to avoid this combination but, in the mean time, the proposed check is good enough, and so I will mark this as resolved. Happy to discuss if I misunderstood :-)

PR #41, opened 2024-05-15
40 metadata writer on sdmx ml

javihern98:
This pull request adds for the first time the Writing on XML feature on pysdmx in Item and ItemScheme classes (and superclasses).

*Summary*:

- Added writer function to introduce a specific dictionary (Message Content, to be implemented afterwards). This is divided in two modules, the public write module and the internal __writer_aux module, for better maintainability and reusability.
- Added Header class with some default values
- Added internal methods _to_XML in base classes

### DIFF
--- a/src/pysdmx/model/message.py
+++ b/src/pysdmx/model/message.py
@@ -1,0 +1,32 @@
+"""Message module.
+
+This module contains the enumeration for the different types of messages that
+can be written.
+"""
+
+from enum import Enum
+
+
+class MessageType(Enum):
+    """MessageType enumeration.
+
+    Enumeration that withholds the Message type for writing purposes.
+    """
+
+    GenericDataSet = 1
+    StructureSpecificDataSet = 2
+    Metadata = 3
+    Error = 4
+    Submission = 5
+
+
+class ActionType(Enum):
+    """ActionType enumeration.
+
+    Enumeration that withholds the Action type for writing purposes.
+    """
+
+    Append = "append"
+    Replace = "replace"
+    Delete = "delete"
+    Information = "information"
--- a/src/pysdmx/model/__base.py
+++ b/src/pysdmx/model/__base.py
@@ -1,9 +1,15 @@
 from datetime import datetime
-from typing import Any, Dict, Optional, Sequence, Union
+from typing import Any, Dict, List, Optional, Sequence, Union
 
 from msgspec import Struct
 
 from pysdmx.errors import ClientError
+from pysdmx.writers.__write_aux import (
+    ABBR_COM,
+    ABBR_STR,
+    add_indent,
+    export_intern_data,
+)
 
 
 class Annotation(Struct, frozen=True, omit_defaults=True):
@@ -57,6 +63,14 @@ class Annotation(Struct, frozen=True, omit_defaults=True):
         return ", ".join(out)
 
 
+ANNOTATION_WRITER = {
+    "title": "AnnotationTitle",
+    "type": "AnnotationType",
+    "url": "AnnotationURL",
+    "text": "AnnotationText",
+}
+
+
 class AnnotableArtefact(Struct, frozen=True, omit_defaults=True, kw_only=True):
     """Annotable Artefact class.
 
@@ -86,6 +100,37 @@ class AnnotableArtefact(Struct, frozen=True, omit_defaults=True, kw_only=True):
                 out.append(f"{k}={str(v)}")
         return ", ".join(out)
 
+    def _to_XML(self, indent: str) -> Any:
+
+        if len(self.annotations) == 0:
+            return ""
+
+        child1 = indent
+        child2 = add_indent(child1)
+
+        outfile = f"<{ABBR_COM}:Annotations>"
+        for annotation in self.annotations:
+            if annotation.id is None:
+                outfile += f"{child1}<{ABBR_COM}:Annotation>"
+            else:
+                outfile += (
+                    f"{child1}<{ABBR_COM}:Annotation " f"id={annotation.id!r}>"
+                )
+
+            for attr, label in ANNOTATION_WRITER.items():
+                if getattr(annotation, attr, None) is not None:
+                    value = getattr(annotation, attr)
+                    value = value.replace("&", "&amp;").rstrip()
+                    outfile += (
+                        f"{child2}<{ABBR_COM}:{label}>"
+                        f"{value}"
+                        f"</{ABBR_COM}:{label}>"
+                    )
+
+            outfile += f"{child1}</{ABBR_COM}:Annotation>"
+        outfile += f"</{ABBR_COM}:Annotations>"
+        return outfile
+
 
 class IdentifiableArtefact(AnnotableArtefact, frozen=True, omit_defaults=True):
     """Identifiable Artefact class.
@@ -103,6 +148,25 @@ class IdentifiableArtefact(AnnotableArtefact, frozen=True, omit_defaults=True):
     uri: Optional[str] = None
     urn: Optional[str] = None
 
+    def _to_XML(self, indent: str) -> Dict[str, Any]:
+        attributes = ""
+
+        if self.id is not None:
+            attributes += f" id={self.id!r}"
+
+        if self.uri is not None:
+            attributes += f" uri={self.uri!r}"
+
+        if self.urn is not None:
+            attributes += f" urn={self.urn!r}"
+
+        outfile = {
+            "Annotations": super(IdentifiableArtefact, self)._to_XML(indent),
+            "Attributes": attributes,
+        }
+
+        return outfile
+
 
 class NameableArtefact(IdentifiableArtefact, frozen=True, omit_defaults=True):
     """Nameable Artefact class.
@@ -116,6 +180,31 @@ class NameableArtefact(IdentifiableArtefact, frozen=True, omit_defaults=True):
 
     name: Optional[str] = None
     description: Optional[str] = None
+
+    def _to_XML(self, indent: str) -> Dict[str, Any]:
+        outfile = super(NameableArtefact, self)._to_XML(indent)
+
+        if self.name is not None:
+            outfile["Name"] = [
+                (
+                    f"{add_indent(indent)}"
+                    f'<{ABBR_COM}:Name xml:lang="en">'
+                    f"{self.name}"
+                    f"</{ABBR_COM}:Name>"
+                )
+            ]
+
+        if self.description is not None:
+            outfile["Description"] = [
+                (
+                    f"{add_indent(indent)}"
+                    f'<{ABBR_COM}:Description xml:lang="en">'
+                    f"{self.description}"
+                    f"</{ABBR_COM}:Description>"
+                )
+            ]
+
+        return outfile
 
 
 class VersionableArtefact(NameableArtefact, frozen=True, omit_defaults=True):
@@ -133,6 +222,22 @@ class VersionableArtefact(NameableArtefact, frozen=True, omit_defaults=True):
     valid_from: Optional[datetime] = None
     valid_to: Optional[datetime] = None
 
+    def _to_XML(self, indent: str) -> Dict[str, List[str]]:
+        outfile = super(VersionableArtefact, self)._to_XML(indent)
+
+        if self.version is not None:
+            outfile["Attributes"] += f" version={self.version!r}"
+
+        if self.valid_from is not None:
+            valid_from_str = self.valid_from.strftime("%Y-%m-%dT%H:%M:%S")
+            outfile["Attributes"] += f" validFrom={valid_from_str!r}"
+
+        if self.valid_to is not None:
+            valid_to_str = self.valid_to.strftime("%Y-%m-%dT%H:%M:%S")
+            outfile["Attributes"] += f" validTo={valid_to_str!r}"
+
+        return outfile
+
 
 class Item(NameableArtefact, frozen=True, omit_defaults=True):
     """Item class.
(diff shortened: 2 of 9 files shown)


src/pysdmx/model/message.py
```diff
@@ -30,3 +35,24 @@ class ActionType(Enum):
...
     Information = "information"
+
+
+class Header(Struct, frozen=True, kw_only=True):
+    """Header for the SDMX-ML file."""
+
+    id: str = "test"
+    test: str = "true"
+    prepared: datetime = datetime.strptime("2021-01-01", "%Y-%m-%d")
+    sender: str = "Unknown"
```
sosna, line 46: There is a convention to use ZZZ whenever the sender is unknown. It is coming from the IMF initially, but it is fairly widespread, as far as I know.

src/pysdmx/model/message.py
```diff
@@ -30,3 +35,24 @@ class ActionType(Enum):
...
+    receiver: str = "Not_Supplied"
+    source: str = "PySDMX"
+    dataset_action: str = ActionType.Information.value
```
sosna, line 49: Receiver, source and dataset action are all with a minOccurs=0 in the SDMX-ML spec. Therefore, we could make them optional and set their default value to None?

src/pysdmx/model/message.py
```diff
@@ -30,3 +35,24 @@ class ActionType(Enum):
...
+
+    def __post_init__(self) -> None:
```
sosna, line 51: This check is no longer necessary if test becomes a boolean?

src/pysdmx/model/message.py
```diff
@@ -4,8 +4,13 @@
...
 
+from datetime import datetime
 from enum import Enum
 
+from msgspec import Struct
+
+from pysdmx.errors import ClientError
+
 
 class MessageType(Enum):
```
sosna, line 15: The name might be a bit confusing as, with MessageType, some might understand Structure, Data, ReferenceMetadata, etc. Or even something more specific like StructureSpecificData2.1, GenericData2.0, etc., etc. Maybe we could rename it to SdmxAction?

src/pysdmx/io/xml/sdmx_two_one/__init__.py
```diff
@@ -0,0 +1 @@
+"""SDMX 2.1 XML reader and writer."""
```
sosna, line 1: This is more a matter of taste, but I tend to prefer shorter package names, whenever possible. An alternative name could be `sdmx21`. This is more a matter of taste, obviously, so feel free to ignore 😊

src/pysdmx/io/xml/sdmx_two_one/writer/__write_aux.py
```diff
@@ -0,0 +1,230 @@
...
+
+from collections import OrderedDict
+from typing import Any, Dict
+
+from pysdmx.model.message import Header, MessageType
+
+MESSAGE_TYPE_MAPPING = {
+    MessageType.GenericDataSet: "GenericData",
+    MessageType.StructureSpecificDataSet: "StructureSpecificData",
+    MessageType.Metadata: "Structure",
```
sosna, line 11: This could be renamed to MessageType.Structure. I mention this because there are also reference metadata messages, which is what people might be thinking about when seeing MessageType.Metadata.

src/pysdmx/io/xml/sdmx_two_one/writer/__write_aux.py
```diff
@@ -0,0 +1,230 @@
...
+        A string with the namespaces
+    """
+    # if type_ == MessageType.GenericDataSet:
+    #     return f"xmlns:{ABBR_GEN}={NAMESPACES[ABBR_GEN]!r} "
+    # elif type_ == MessageType.StructureSpecificDataSet:
+    #     return f"xmlns:{ABBR_SPE}={NAMESPACES[ABBR_SPE]!r} "
+    # elif type_ == MessageType.Metadata:
+    #     return f"xmlns:{ABBR_STR}={NAMESPACES[ABBR_STR]!r} "
+    # else:
+    #     return ""
```
sosna, line 60: Should we remove code that is commented out?

src/pysdmx/io/xml/sdmx_two_one/writer/__write_aux.py
```diff
@@ -0,0 +1,230 @@
...
+    return f"xmlns:{ABBR_STR}={NAMESPACES[ABBR_STR]!r} "
+
+
+# def __namespaces_from_content(content: Dict[str, Any]) -> str:
```
sosna, line 64: Same comment as above

src/pysdmx/io/xml/sdmx_two_one/writer/__write_aux.py
```diff
@@ -0,0 +1,230 @@
...
+#     """Returns the namespaces for the XML file based on content.
```
sosna, line 65: Same as above

src/pysdmx/io/xml/sdmx_two_one/writer/metadata_writer.py
```diff
@@ -0,0 +1,276 @@
+"""Module for writing metadata to XML files."""
```
sosna, line 1: Similar comments about 1) long package names and 1) the fact that metadata might mean reference metadata to some people.

So, instead of:
`pysdmx.io.xml.sdmx_two_one.writer.metadata_writer`
we could have:
`pysdmx.io.xml.sdmx21.writer.structure`

src/pysdmx/io/xml/sdmx_two_one/writer/metadata_writer.py
```diff
@@ -0,0 +1,276 @@
...
+
+    for key in MSG_CONTENT_PKG:
+        outfile += __write_metadata_element(content, key, prettyprint)
+
+    outfile += f"{nl}{child1}</{ABBR_MSG}:Structures>"
+
+    return outfile
+
+
+def export_intern_data(data: Dict[str, Any], indent: str) -> str:
```
sosna, line 245: Apparently, only `generate_structure` is imported? If so, should this method be public?

src/pysdmx/io/xml/sdmx_two_one/writer/metadata_writer.py
```diff
@@ -0,0 +1,276 @@
...
+        The XML string with the exported data
+    """
+    outfile = get_outfile(data, "Annotations", indent)
+    outfile += get_outfile(data, "Name", indent)
+    outfile += get_outfile(data, "Description", indent)
+
+    return outfile
+
+
+def get_outfile(obj_: Dict[str, Any], key: str = "", indent: str = "") -> str:
```
sosna, line 262: Same as above

src/pysdmx/io/xml/sdmx_two_one/__init__.py
javihern98: Changed to sdmx21

src/pysdmx/io/xml/sdmx_two_one/writer/__write_aux.py
javihern98: Changing the MessageType to Structure.

src/pysdmx/io/xml/sdmx_two_one/writer/metadata_writer.py
javihern98: Changing it to structure, giving the new file structure makes more sense.

src/pysdmx/io/xml/sdmx_two_one/writer/metadata_writer.py
javihern98: Yes, changing the rest of methods to private (with __ at the beginning)

src/pysdmx/model/message.py
```diff
@@ -30,3 +35,24 @@ class ActionType(Enum):
     Replace = "replace"
     Delete = "delete"
     Information = "information"
+
+
+class Header(Struct, frozen=True, kw_only=True):
+    """Header for the SDMX-ML file."""
```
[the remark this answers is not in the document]
javihern98: Changed the Docstring

src/pysdmx/model/message.py
javihern98: We could change it, but never saw this ZZZ sender. I will make it as the default value.

src/pysdmx/model/message.py
javihern98: Removing this check as set test to boolean.

src/pysdmx/model/message.py
javihern98: @sosna Need another name different than SdmxAction, it has to be some name that refers to the type of message we want to write (data, structures, etc)

src/pysdmx/model/message.py
```diff
@@ -30,3 +35,24 @@ class ActionType(Enum):
...
+
+    id: str = "test"
+    test: str = "true"
```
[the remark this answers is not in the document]
javihern98: Changed to Boolean

src/pysdmx/model/message.py
javihern98: Yes, also adding a post_init check to ensure the dataset_action is one of the ActionType values.

javihern98: Edited as optional on all values and set type of dataset_action as Optional[ActionType]

src/pysdmx/model/message.py
javihern98: The Message class should, theoretically, be able to write also Reference Metadata or any type of SDMX classes, including Submission Results and Errors (from FMR and SDMX APIs). I would prefer to keep MessageType or try to find a synonym that explains better this capability.

sosna: Thanks, Javier. It's all clear now and, based on what you described, I agree the name is a good one 😊.

However, as it seems to be strongly related to writing messages (which I didn't know when making my comment), maybe it could be moved below the io module, or even io.sdmx21, as it seems to be SDMX-ML related?

javihern98: Yes, makes sense to move it to io.xml at least for now, I cannot think of any other place where this could be used, AFAIK the other flavours of SDMX does not have any of these values (support for Submission and Error messages, differentiation between Generic and Structure Specific...). The value of this Enum must be specified when writing only to SDMX ML files.

javihern98: Moved MessageType to io.xml.enums.py

src/pysdmx/model/message.py
```diff
@@ -0,0 +1,35 @@
...
+    Delete = "delete"
+    Information = "information"
+
+
+class Header(Struct, frozen=True, kw_only=True):
+    """Header for the SDMX messages."""
+
+    id: str = "test"
+    test: bool = True
+    prepared: datetime = datetime.strptime("2021-01-01", "%Y-%m-%d")
```
sosna, line 31: Sorry for not spotting this earlier but, I think there are 2 issues with this:
- It should be a time stamp (i.e. date and time), not a date.
- Maybe the date should not be hardcoded, but should instead be the time when the Header gets instantiated, in the UTC timezone?

src/pysdmx/model/message.py
```diff
@@ -0,0 +1,35 @@
...
+    Append = "append"
+    Replace = "replace"
+    Delete = "delete"
+    Information = "information"
+
+
+class Header(Struct, frozen=True, kw_only=True):
+    """Header for the SDMX messages."""
+
+    id: str = "test"
```
sosna, line 29: Sorry for not spotting this earlier: Instead of hardcoding the ID, maybe with could use the Python uuid module, and its uuid4 function? After all, the ID is expected to be unique for each message so, that would be a good default unless the caller sets the desired ID?

src/pysdmx/model/message.py
javihern98: My best bet would be to use datetime.now() as default value, but is a nightmare to get the 100% code coverage. I will have another look at it

src/pysdmx/model/message.py
javihern98: Same answer as with the prepared default value. I like it, but need to find a way to properly test it. I will have another look at it.

javihern98: Changed to str(uuid4()) on latest commit.

src/pysdmx/model/message.py
javihern98: Changed to datetime.now() on latest commit.

src/pysdmx/model/message.py
```diff
@@ -0,0 +1,36 @@
...
+    Delete = "delete"
+    Information = "information"
+
+
+class Header(Struct, frozen=True, kw_only=True):
+    """Header for the SDMX messages."""
+
+    id: str = str(uuid.uuid4())
+    test: bool = True
+    prepared: datetime = datetime.now()
```
sosna, line 32: Thanks!. As mentioned, I would avoid "naive" datetime objects by specifying a default time zone of UTC. As mentioned in the Python docs for the datetime module: "Warning: Because naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC is by calling datetime.now(timezone.utc)"

The rest looks good to me, thanks a lot :)

javihern98: Okay, adding it, every day you learn something new 😄

sosna: And this never stops, in IT 😂

javihern98:
Ready for final review @sosna @stratosn

sosna (approved):
Thanks, @javihern98!

PR #46, opened 2024-05-17
44 read xml parsers and add lxml dependency

javihern98:
This feature adds the XML reading compatibility.

Summary:

- Added read_xml method and optional (but default) validation against XSD files in SDMX-ML 2.1
- Added input processor to differentiate between JSON and XML files
- Added SubmissionResult and SDMX Error messages compatibility.
- Added lxml and xmltodict dependency.


javihern98:
Seems that CI/CD testing is not installing the optional dependency xmltodict

.gitignore
```diff
@@ -43,3 +43,6 @@ coverage-report.xml
 qa/
 dist/
 .env
+
+# Ignore poetry lock file
+poetry.lock
```
sosna, line 48: We should not do this. This will make the developers environments, as well as CI/CD pipelines non deterministic: https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control

poetry.lock
```diff
@@ -1,1543 +0,0 @@
-# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
sosna, line 1: This file should not be removed (see previous comment)

src/pysdmx/util/__init__.py
```diff
@@ -25,6 +25,11 @@ class Reference(Struct, frozen=True):
     id: str
     version: str
 
+    @property
+    def full_id(self) -> str:
```
sosna, line 29: I would be a bit careful with introducing new ways to reference artefacts. SDMX has 2:
- One where we need to use the artefact type, its ID, agency and version;
- One where we can use the SDMX URN.

Do we need a 3rd one?

src/pysdmx/model/submission.py
```diff
@@ -0,0 +1,20 @@
+"""SDMX Submission classes."""
```
sosna, line 1: This class could also go to the message module, IMHO.

src/pysdmx/model/submission.py
```diff
@@ -0,0 +1,20 @@
...
+
+from msgspec import Struct
+
+
+class SubmissionResult(Struct):
```
sosna, line 6: Do we want to make it immutable too?

sosna (commented):
Re. `src/pysdmx/io/xml/sdmx21/schemas`, I still think we would need another mechanism, and have the schemas outside the library package. @stratosn, what do you think?

src/pysdmx/util/__init__.py
javihern98: It is only a convenience method to extract the "full_id", i.e, the combination of agency_id, id and version (don´t know if there is a formal name for it). We could add it to utils

.gitignore
sosna: Yes, indeed, the poetry.lock should include everything, including optional dependencies.

src/pysdmx/util/__init__.py
sosna: That sounds OK to me. It would be nice to find a good name for it. For me, there is the ID and the URN. I also like the "relevant part of the URN", i.e. Type=Agency:Id(Version), as it contains all information while being shorter than the URN. 

But a full ID without the type sounds a bit odd, without knowing how it's used. After all, Codelist=SDMX:FREQ(1.0) and ConceptScheme=SDMX:FREQ(1.0) would have the same `full_id` but would nevertheless be 2 different artefacts...

src/pysdmx/model/submission.py
javihern98: Yes, probably. Should be frozen. I will make the change.

src/pysdmx/model/submission.py
javihern98: No strong opinion on why it should be on a separate file, so moving it.

src/pysdmx/util/__init__.py
javihern98: Could we then, as Reference is defined also with sdmx_type, make the __str__ method as sdmxtype=AGENCY_ID:ID(VERSION)? This should cover all use cases.

javihern98: And only call it Reference, I cannot find a better way to define it.

sosna: Yes, if you are happy with adding the type too, then, this is perfect, thanks :). This then is the "relevant part of the URN" I was mentioning previously.

We could return it via a `__str__` method, as you suggested, which is nice, I think 😊. Or we can have a property, but, instead of calling it `full_id`, I would call it something like `short_urn`. 

But I think I prefer your suggestion of using `__str__`

javihern98: Implementing it as __str__. On Reference, should we add a post_init check for the sdmx_type? What should be the values that sdmx_type may take? Is there any list of those values?

sosna: Thanks, @javihern98, this is nice :)

Re. your question, we could add this. As you may remember, I defined such an enum in the Structure query builder. Once it's ready, I could add it to the model instead. But, for the time being, I guess it's OK like that?

sosna (commented):
As discussed, I think there is an overall agreement to have an external package with all the schemas (all versions, all formats, maintained by the TWG, although we can create the initial package, in order to not be blocked)?

sosna (changes requested):
As discussed, I think there is an overall agreement to have an external package with all the schemas (all versions, all formats, maintained by the TWG, although we can create the initial package, in order to not be blocked)?

javihern98:
Sure, let´s generate the new package with the schemas. Any desired name? @sosna

sosna:
I'd say something like sdmx-schemas?
We should then have folders per version (1.0, 2.0, 2.1 and 3.0), and then formats (json, xml)

Maybe the schemas could be compressed?

Also, do we need any Python code, like functions you call to return the appropriate schema?

Once we have clarified what we want, we could decide who does what?

javihern98:
Generated new schemas library here: https://pypi.org/project/sdmxschemas/

Updated code to import it and use the SDMX 2.1 schemas.

stratosn (approved):

sosna (approved):
Thanks, @javihern98. Great work 😊!

PR #47, opened 2024-05-17
Added message class with the actual content implemented (ItemScheme related)

### LINKED ISSUE #42: The Message class
Implement Message class based on MsgSpec/PyDantic as specified in SDMXthon.

https://docs.sdmxthon.meaningfuldata.eu/packages/model/message.html

javihern98:
This PR adds the Message class to the library, which allows the management of different SDMX objects at once, to enable the writing of complex SDMX messages.

Summary:
- Added the Message class and related tests

Closes #42

### DIFF
--- a/src/pysdmx/model/message.py
+++ b/src/pysdmx/model/message.py
@@ -11,6 +11,7 @@ import uuid
 
 from msgspec import Struct
 
+from pysdmx.errors import ClientError
 from pysdmx.model import Codelist, ConceptScheme
 from pysdmx.model.__base import ItemScheme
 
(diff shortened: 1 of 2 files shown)


javihern98:
@sosna Ready for review. The conflicts will be solved locally once you approve

src/pysdmx/model/message.py
```diff
@@ -34,3 +37,80 @@ class Header(Struct, frozen=True, kw_only=True):
...
+
+    def __get_element_by_uid(self, type_: str, unique_id: str) -> Any:
+        """Returns a specific element from content."""
+        if type_ not in self.content:
+            raise ValueError(f"No {type_} found")
+
+        if unique_id in self.content[type_]:
+            return self.content[type_][unique_id]
+
+        raise ValueError(f"No {type_} with id {unique_id} found in content")
```
sosna, line 92: I don't know how this is being used but, if not having something of a certain type is acceptable, we could return an empty collection, instead of throwing an error. 

If it is not acceptable and so an error must be returned, then maybe we should return the NotFound error provided by the pysdmx `errors` module?

src/pysdmx/model/message.py
```diff
@@ -34,3 +37,80 @@ class Header(Struct, frozen=True, kw_only=True):
...
+            ``Codelist``, etc.).
+    """
+
+    content: Dict[str, Any]
+
+    def __post_init__(self) -> None:
+        """Checks if the content is valid."""
+        for content_key, content_value in self.content.items():
+            if content_key not in MSG_CONTENT_PKG:
+                raise ValueError(f"Invalid content type: {content_key}")
```
sosna, line 69: Here as well, in case something is invalid, we could return the ClientError in the pysdmx `errors` module

src/pysdmx/model/message.py
javihern98: The goal is to provide a method to extract a specific element from the Message content. I do not have a strong option to return an empty dictionary or an Error. I do prefer an error for better usability.

I will implement it using the NotFound error, do you agree? @sosna

src/pysdmx/model/message.py
javihern98: Yes, I will implement it using the ClientError exception.

src/pysdmx/model/message.py
sosna: Sure, that sounds good to me. Thanks, @javihern98

sosna (approved):

PR #58, opened 2024-06-28
Add query builder for structure queries

sosna:
This PR adds a query builder for structure queries, as agreed a few weeks ago.


javihern98 (approved):
This is a much better implementation that what we currently have on sdmxthon (and more maintainable). 😄 

No more comments to add after reviewing it in depth. I cannot check the standard against all these enums due to lack of time. We have plenty of tests to query the actual REST APIs of the SDMX agencies that may be added in a latter issue.

sosna:
@javihern98, thanks a lot! Very happy to hear you like the implementation 😊! Yes, any additional test is always welcome (we can never have enough tests 😊). I hope I can send a PR for the schema queries towards the end of the week.

Issue #49, opened 2024-05-21
CSV parsers for data reading and writing


javihern98:
- Add CSV parsers for SDMX versions 1.0 and 2.0
- Add CSV writers for SDMX versions 1.0 and 2.0
- Add pandas dependency

PR #59, opened 2024-07-01
50 xml parsers for data reading

javihern98:
Added data parsers for SDMX-ML 2.1 files.

This PR adds a dummy class DataSet that will be modified afterwards, once agreed on the functionality that this class will provide, but currently it is necessary to ensure the basic elements are correctly implemented.


javihern98:
This PR includes a temprary dataset class to be replaced once CSV parsers are also included and discussions have been made, just to ensure all functionality is consistent to what we have in SDMXthon

src/pysdmx/util/handlers.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Returns:
+        full id.
+    """
+    full_id = obj_.split("=", 1)[1]
+    return split_unique_id(full_id)
+
+
+def add_list(element: Any) -> Any:
+    """Adds data contained in the xml Dataset into a list if it is possible.
```
sosna, line 42: I guess this is not xml specific, isn't it (considering its location in the package)? If so, I'd remove xml from the docstring. It's also hard for me to understand what this is about and why it should be used. Maybe some more documentation could be added? Else, if this is io-specific only, we could also have a util module below the io module. As you wish :)

src/pysdmx/util/handlers.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Returns:
+        The elements in a list.
+    """
+    if not isinstance(element, list):
+        element = [element]
+    return element
+
+
+# def drop_na_all(df: pd.DataFrame):
```
sosna, line 55: Should we drop commented out code and add it back when we are ready?

src/pysdmx/util/handlers.py
```diff
@@ -0,0 +1,59 @@
...
+
+    Args:
+        obj_: String with the name of the object.
+        split_id: True splits the id.
+
+    Returns:
+        full id.
+    """
+    full_id = obj_.split("=", 1)[1]
+    return split_unique_id(full_id)
```
sosna, line 38: These 2 functions sound similar to me to the parse_urn function in this file: https://github.com/bis-med-it/pysdmx/blob/develop/src/pysdmx/util/__init__.py?

src/pysdmx/io/xml/sdmx21/reader/data_read.py
```diff
@@ -0,0 +1,342 @@
...
+    URN,
+    VALUE,
+    VERSION,
+)
+from pysdmx.util.handlers import add_list, split_from_urn
+
+chunksize = 50000
+
+
+class Dataset:
```
sosna, line 37: I guess that, eventually, we'll want to create a Dataset class under pysdmx.model, isn't it? And, like all other model classes, it should be format-agnostic? If so, I think we can go ahead with this class now but we should discuss about the dataset class and create it before the 1.0.0 release. What do you think?

sosna: Sorry, forget about the above. I see you already mentioned the same in your PR :)

src/pysdmx/util/handlers.py
javihern98: Yes, this is due to how xmltodict processes the XML, if there is only one child element, it creates a dictionary, but if there is more than one child element, it creates a list of dictionaries. 

This is io specific only, and specific only for xml, so I will move it to io.xml.utils.py and update the docstring

src/pysdmx/util/handlers.py
javihern98: Sure, I will remove it

src/pysdmx/util/handlers.py
javihern98: I will recheck it, these are functions that come directly from sdmxthon so probably I missed something. The parse_urn function worked fine for metadata so it should also work here in theory.

src/pysdmx/io/xml/sdmx21/reader/data_read.py
javihern98: Yes, closing this as it will be modified in #49 and some discussions are pending.

src/pysdmx/util/handlers.py
javihern98: Solved in latest commit.

src/pysdmx/util/handlers.py
javihern98: Solved in latest commit and used parse_urn.

sosna (approved):
Great, thanks a lot, @javihern98

PR #62, opened 2024-07-09
Query builders for reference metadata queries

sosna:
This PR adds support for the 3 types of reference metadata queries supported in the SDMX-REST API.


javihern98 (approved):
All good on our side, this part of the standard (Reference Metadata on Metadataflows) is not related to anything we have ever implemented, so review is based only on code structure and test quality, which seems really nice :)

sosna:
Thanks, @javihern98. The builder for data queries is coming next :)

PR #65, opened 2024-07-17
49 csv parsers for data reading

javihern98:


src/pysdmx/model/message.py
```diff
@@ -36,6 +36,14 @@ class ActionType(Enum):
     Information = "information"
 
 
+SDMX_CSV_ACTION_MAPPER = {
```
sosna, line 39: This is format-specific. Therefore, I would suggest to move this into its respective io package.

src/pysdmx/model/dataset.py
```diff
@@ -0,0 +1,22 @@
+"""Dataset module."""
+
+from typing import Any, Dict
+
+from msgspec import Struct
+import pandas as pd
+
+
+class Dataset(Struct, frozen=False, kw_only=True):
```
sosna, line 9: I have several comments about this Dataset class but let's discuss them in the afternoon, as agreed :).

src/pysdmx/io/csv/sdmx10/reader/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    # Get Dataframe from CSV file
+    df_csv = pd.read_csv(StringIO(infile))
+    # Drop empty columns
+    df_csv = df_csv.dropna(axis=1, how="all")
+
+    # Determine SDMX-CSV version based on column names
+    if "DATAFLOW" not in df_csv.columns:
+        # Raise an exception if the CSV file is not in SDMX-CSV format
+        raise Exception("Invalid CSV file, only SDMX-CSV is allowed")
```
sosna, line 50: I think we should use pysdmx errors instead (cf. pysdmx/errors.py)?

src/pysdmx/io/csv/sdmx20/reader/__init__.py
```diff
@@ -0,0 +1,91 @@
...
+    # Drop empty columns
+    df_csv = df_csv.dropna(axis=1, how="all")
+
+    # Determine SDMX-CSV version based on column names
+    if (
+        "STRUCTURE" not in df_csv.columns
+        or "STRUCTURE_ID" not in df_csv.columns
+    ):
+        # Raise an exception if the CSV file is not in SDMX-CSV format
+        raise Exception("Invalid CSV file, only SDMX-CSV is allowed")
```
sosna, line 56: Same comment about the need to use pymedal errors.

sosna (commented):
It looks nice, thanks :).

Though, as agreed, we need to discuss about the Dataset class

src/pysdmx/model/message.py
javihern98: Moving it to io.csv.sdmx20

src/pysdmx/model/dataset.py
javihern98: Agreed on meeting #72

src/pysdmx/io/csv/sdmx10/reader/__init__.py
javihern98: Reviewing exceptions and changing them to ClientError

javihern98:
> It looks nice, thanks :).
> 
> Though, as agreed, we need to discuss about the Dataset class

After further discussion, we decided that:

- The Dataset class will be renamed to PandasDataset, with the following properties:
  - SDMX Dataset related properties (action, reporting_begin, etc...)
  - data will be hold as Pandas Dataframe in this class
  - structure will have the URN of the Schema referenced, or the actual Schema instance of the DSD, DataFlow or Provision Agreement
  - short_urn: gets the meaningful part of the URN to reference the schema, in the format of "STRUCTURE_TYPE=AGENCY_ID:ID(VERSION)". Designed as a convenience method to simply identify the dataset in an unique way
  - attributes: withholds the information of the attributes at dataset level

@sosna Please comment if I missed something

javihern98:
Summary of changes:

- Refactored Dataset (as specified in discussion: https://github.com/bis-med-it/pysdmx/pull/65#issuecomment-2257654784)
- Added Action handling and two + actions in the same file (raise error, check if this behavior is correct)
- Added support for Provision Agreement
- Use of full URN in dataset loading
- Removed XML dataset classes and unify functionalities
- Added dedicated tests for several different structures, handling of short_urn and specific Client Errors

javihern98:
@sosna Even after removing the test with the big file seems I am still getting segmentation fault errors on Ubuntu, any idea?

sosna (approved):

PR #73, opened 2024-07-22
Use the new query builders in the fmr module

sosna:


javihern98 (approved):
All changes reviewed 👍 

Is there any possibility to implement a method that checks any query generated to ensure we do not generate querys with //?

sosna:
> Is there any possibility to implement a method that checks any query generated to ensure we do not generate queries with //?

Thanks, @javihern98. Do you happen to have an example of this behavior? Because, I don't expect this to happen, neither with the full URL (with default values), nor with the short one, and so if you have observed this, it must be a bug.

javihern98:
> > Is there any possibility to implement a method that checks any query generated to ensure we do not generate queries with //?
> 
> Thanks, @javihern98. Do you happen to have an example of this behavior? Because, I don't expect this to happen, neither with the full URL (with default values), nor with the short one, and so if you have observed this, it must be a bug.

Not that I found, I do not expect it either, but just to double check and be sure about it.

sosna:
> Not that I found, I do not expect it either, but just to double check and be sure about it.

OK, thanks, I'll keep that in mind in my upcoming tests (I will add a few more tests and will also build a first connector)

PR #85, opened 2024-09-24
Move SeriesInfo to dataset module

sosna:
This is so as to keep the data-related classes in the same module


javihern98 (approved):
All good 😄

PR #94, opened 2024-11-06
Fix issues with deserialization of SDMX-JSON output

### LINKED ISSUE #93: Fix issues with SDMX-JSON deserialization
There were several discrepancies between what the FMR used to generate in SDMX-JSON, and the official SDMX-JSON schemas. For example:

- The schema for provision agreements has a property called [dataflow](https://github.com/sdmx-twg/sdmx-json/blob/master/structure-message/tools/schemas/2.0.0/sdmx-json-structure-schema.json#L4688). The FMR called it `structureUsage` instead.
- According to the spec, there can only be one [time dimension](https://github.com/sdmx-twg/sdmx-json/blob/master/structure-message/tools/schemas/2.0.0/sdmx-json-structure-schema.json#L3539). So, it is not an array, and it is spelled `timeDimension`. However, in the FMR output, it wass an array and the property was called `timeDimensions`.
- In SDMX-JSON structure messages, `CodedTextFormatType` has a property called [dataType](https://github.com/sdmx-twg/sdmx-json/blob/master/structure-message/tools/schemas/2.0.0/sdmx-json-structure-schema.json#L2515). Same for [BasicComponentTextFormatType](https://github.com/sdmx-twg/sdmx-json/blob/master/structure-message/tools/schemas/2.0.0/sdmx-json-structure-schema.json#L2297). However, in the SDMX-JSON output generated by the FMR, this property was called `textType` instead.

These issues have been fixed in the latest versions of the FMR but pysdmx deserializers for SDMX-JSON are still based on the old (and wrong) output...

This must be fixed before the first official release (1.0.0).

sosna:
Fix #93

### DIFF
--- a/tests/api/fmr/samples/df/hierarchy_schema.json
+++ b/tests/api/fmr/samples/df/hierarchy_schema.json
@@ -395,27 +395,25 @@
                                 "conceptIdentity": "urn:sdmx:org.sdmx.infomodel.conceptscheme.Concept=BIS:CONCEPTS(1.0).OPTION_TYPE"
                             }
                         ],
-                        "timeDimensions": [
-                            {
-                                "position": 4,
-                                "type": "TimeDimension",
-                                "links": [
-                                    {
-                                        "rel": "self",
-                                        "type": "timedimension",
-                                        "uri": "https://raw.githubusercontent.com/sdmx-twg/sdmx-json/develop/structure-message/tools/schemas/2.0.0/sdmx-json-structure-schema.json",
-                                        "urn": "urn:sdmx:org.sdmx.infomodel.datastructure.TimeDimension=BIS.TEST(1.0).TIME_PERIOD"
-                                    }
-                                ],
-                                "id": "TIME_PERIOD",
-                                "localRepresentation": {
-                                    "format": {
-                                        "textType": "ObservationalTimePeriod"
-                                    }
-                                },
-                                "conceptIdentity": "urn:sdmx:org.sdmx.infomodel.conceptscheme.Concept=BIS:CONCEPTS(1.0).TIME_PERIOD"
-                            }
-                        ]
+                        "timeDimension": {
+                            "position": 4,
+                            "type": "TimeDimension",
+                            "links": [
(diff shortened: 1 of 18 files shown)


sosna:
There is a new issue with the build system on MacOS (poetry fails to install a package). I don't know why yet. This is not an issue with the code itself (i.e. I ran the same build on my Mac locally, and it completes successfully). Therefore, I have temporarily disabled the macos builds, so that we can proceed with the release.

I have also added a build against the latest version of Python (3.13).

tests/api/fmr/samples/df/hierarchy_schema.json
```diff
@@ -388,34 +385,30 @@
...
-                                    }
-                                },
-                                "conceptIdentity": "urn:sdmx:org.sdmx.infomodel.conceptscheme.Concept=BIS:CONCEPTS(1.0).TIME_PERIOD"
-                            }
-                        ]
+                        "timeDimension": {
+                            "links": [
+                                {
+                                    "rel": "self",
+                                    "type": "timedimension",
```
stratosn, line 399: If I understand correctly, there is no `type` property in the definition of the `TimeDimension` object.

sosna: Indeed. Same for the Dimension object. In addition, for timeDimension, `position` is not allowed

stratosn: Hopefully, when we get stricter json schemas, we may be able to resolve such cases :-)

javihern98 (approved):
All good from our side ☺️

javihern98:
Regarding the macOS build issue, we had similar problems with other projects, seems to be an issue with a dependency of poetry.

I suggest using pyproject-hooks==1.0.0 as a dev package and try again. This is not only a problem related with macOS as we had similar issues on ubuntu.

Another solution would be to add some code to the github actions to upgrade the pip, setuptools and wheel versions, maybe they are outdated, in any case we need to dive into it for the final 1.0.0 version.

PR #96, opened 2024-11-11
Set version to 1.0.0-rc-1

sosna:


stratosn (approved):
Exceptionally approved, due to minimal impact

PR #102, opened 2024-11-13
Component is an array only if maxOccurs > 1

### LINKED ISSUE #100: array_def requires additional checks
The following snippet is not sufficient to determine whether a component is an array or not:

```python
def to_array_def(self) -> Optional[ArrayBoundaries]:
        """Returns the array boundaries, if any."""
        if self.minOccurs or self.maxOccurs:
            m = self.minOccurs if self.minOccurs is not None else 0
            return ArrayBoundaries(m, self.maxOccurs)
        else:
            return None
```

We should return an ArrayBoundary only if self.maxOccurs is set to a value greater than 1.

sosna:
Fix #100


javihern98 (approved):
All good from our side ☺️

PR #106, opened 2024-11-14
Fix dependency issues (parsy and pandas)

### LINKED ISSUE #97: Dependency issues
The `fmr` extra seems to required additional libraries (`parsy` and `pandas`). This is not as expected and should be investigated.

sosna:
Fix #97


javihern98 (approved):
All good, nice solution for Pandas dependency and Pandas Dataset☺️ . Shall we add some kind of error to state that specific packages of the library requires an extra to be installed?

sosna:
This would be really nice! Maybe something we can add to the backlog, as improvement for later? We could check if a package is available on the path and, if not, have a nice error message indicating the required extra, as you mentioned.

PR #114, opened 2024-11-19
Pass both title and description (if available) to Python error class

### LINKED ISSUE #112: Main error should pass both title and description
something like super().__init__(f"{title}: {description}") but we also need to check for nulls

sosna:
Fix #112


javihern98 (approved):
All good, nice improvement! 😊

sosna:
Thanks, @javihern98. Glad you like it 😊

PR #115, opened 2024-11-19
Replace deprecated utcnow()

### LINKED ISSUE #111: Stop using utcnow
We should stop using utcnow, as it returns a naïve datetime object

sosna:
Fix #111


javihern98 (approved):
All good 😊

PR #119, opened 2024-11-19
Handle case where a component has both a local and a core enumerated representation

### LINKED ISSUE #108: Problem with parsing Components that have different Enumeration that their Concept's core Enumeration
Assuming a DSD with a coded Component (eg an Attribute with local Representation `CL_OBS_STATUS(2.0)`), which uses a Concept with a different coding (core Representation, eg `CL_OBS_STATUS(1.0)`), then parsing has a problem when building the schema.

The schema query will return all Concepts and Codelists used by the DSD, but not the Codelists referenced by the Concepts in their Concept Schemes. So it it will include `CL_OBS_STATUS(2.0)` but **not** `CL_OBS_STATUS(1.0)`

So when parsing the Components, during the instantiation of a Component, `pysdmx` tries to get the model of the used Concept calling the `to_model` method, which tries to get the Codes of that Concept. Since that core Representation refers to a Codelist different than the one used by the Component in the DSD, then the Codelist is not available and this raises an exception.

As an example, this happens in the call, below:
https://github.com/bis-med-it/pysdmx/blob/1cafa99cdef22086cbf53487ce4a78b641665d79/src/pysdmx/io/json/sdmxjson2/messages/dsd.py#L164

sosna:
Fix #108

### DIFF
--- a/src/pysdmx/io/json/fusion/messages/core.py
+++ b/src/pysdmx/io/json/fusion/messages/core.py
@@ -5,6 +5,7 @@ from typing import Any, Optional, Sequence
 
 import msgspec
 
+from pysdmx.errors import NotFound
 from pysdmx.model import ArrayBoundaries, Codelist, Facets
 from pysdmx.util import find_by_urn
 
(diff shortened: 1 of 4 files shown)


stratosn (commented):
All good on my side - leaving open for MD to also check

javihern98 (approved):
All good 😊

PR #121, opened 2024-11-21
Add function to list dataflows available in the Registry

### LINKED ISSUE #90: Add a method to the FMR module to list dataflows
We should add an method to the Registry clients allowing to list the dataflows available in the target.

At the moment, this is only possible via the category scheme...

sosna:
This is based on a proposal from Ole (ECB).

Fix #90


javihern98 (approved):
All good 😊

PR #124, opened 2024-11-22
Review CI pipeline

### LINKED ISSUE #95: Investigate why the build is failing with macos

sosna:
Fix 2 issues with CI pipeline:

- A different version of Python was used by pipx than the one specified by the GitHub action;
- MacOS builds were broken (related to the fact that pipx was using Python 3.13 along with poetry);

Unfortunately, this means that the build against 3.13 cannot be enabled yet...

Fix #95

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
       fail-fast: false
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,12 +23,12 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


stratosn (approved):

stratosn (approved):

.github/workflows/ci.yml
```diff
@@ -16,19 +16,19 @@ jobs:
...
     - name: Checkout code
       uses: actions/checkout@v4
-    - name: Install poetry
-      run: pipx install poetry
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Install poetry
+      run: pipx install --python ${{ matrix.python-version }} poetry
```
javihern98, line 31: Maybe use here pip with --break-system-packages as this is an isolated environment, to avoid issues with pipx and latest python version, as stated in the email.

javihern98 (approved):
All good as python 3.13 is not a priority on our side, in any case please check the above comment.

.github/workflows/ci.yml
sosna: Thanks, @javihern98. Let's proceed as you suggested in your email, while I already merge this in the meantime.

PR #128, opened 2024-11-29
yyyy should be mapped to %Y

### LINKED ISSUE #126: Issue with date pattern for yyyy
`yyyy` date pattern is currently mapped to `%G`. It should be mapped to `%Y` instead.

sosna:
Fix #126


javihern98 (approved):
All good 😊

Issue #92, opened 2024-11-05
Add retrieval of VTL artefacts to the Registry clients


sosna:

sosna:
The following file can be used for testing: [vtl.json](https://github.com/user-attachments/files/17995193/vtl.json)

sosna:
The following file can be used to test [fusion-json](https://github.com/user-attachments/files/18039501/vtl.fusion.json)

sosna:
Additions for the tests.

Fusion-JSON
"mapped": "urn:sdmx:org.sdmx.infomodel.datastructure.Dataflow=FR1:BPE_DETAIL(1.0)",
                    "toVtlMethod": "to_test",
                    "toVtlSubSpace": [
                        "ID_EQUIPEMENT"
                    ],
                    "fromVtlMethod": "from_test",
                    "fromVtlSuperSpace": [
                        "ID_EQUIPEMENT"
                    ]

SDMX-JSON
"dataflow": "urn:sdmx:org.sdmx.infomodel.datastructure.Dataflow=FR1:BPE_DETAIL(1.0)",
                        "fromVtlMapping": {
                            "type": "from_test",
                            "fromVtlSuperSpace": [
                                "ID_EQUIPEMENT"
                            ]
                        },
                        "toVtlMapping": {
                            "type": "to_test",
                            "toVtlSubSpace": [
                                "ID_EQUIPEMENT"
                            ]
                        },
                        "alias": "BPE_DETAIL_VTL"

PR #135, opened 2024-12-03
91 implement the vtl artefacts parsers

### LINKED ISSUE #91: Implement the VTL artefacts parsers
Implement the VTL Artefacts based on SDMX 3.0. Implement XML parsers based on SDMX 2.1 and make a mapping to the new model class.

image

@sosna Shall we use the [vtlengine](https://github.com/Meaningful-Data/vtlengine) to perform syntax/semantic validations?

javihern98:
Fix #91

### DIFF
--- a/src/pysdmx/model/vtl.py
+++ b/src/pysdmx/model/vtl.py
@@ -77,6 +77,16 @@ class Transformation(Item, frozen=True, omit_defaults=True, kw_only=True):
     Result: str
     isPersistent: bool = False
 
+    @property
+    def full_expression(self) -> str:
+        """Return the full expression with the semicolon."""
+        if self.isPersistent:
+            assign_operand = "<-"
+        else:
+            assign_operand = ":="
+
+        return f"{self.Result} {assign_operand} {self.Expression};"
+
     # TODO: Use VTL Engine for syntax/semantic validation based on VTL?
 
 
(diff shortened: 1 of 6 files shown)


javihern98:
Pending finishing #92 and merging the changes

javihern98:
@sosna This is now ready for review after merge from develop with latest changes. I added the check for the VTL Version and some docs on specific artefacts in VTL

src/pysdmx/model/vtl.py
```diff
@@ -88,11 +109,30 @@ class CustomType(Item, frozen=True, omit_defaults=True):
...
+    VTL_2_0 = "2.0"
+    VTL_2_1 = "2.1"
+
+
 class VtlScheme(ItemScheme, frozen=True, omit_defaults=True):
     """A VTL item scheme with the additional propery 'vtl_version'."""
 
     vtl_version: Optional[str] = None
 
+    def __post_init__(self) -> None:
```
sosna, line 125: There are other VTL artefacts which expect a vtl_version, unless I'm mistaken. Should a similar check be added there too?

Also, should vtl_version be an enum then, instead of a str (just a question)?

javihern98: I guess vtl_version could be either a enum or a string, but I do prefer to leave it as string until we find proper use cases. The vtlVersion property is just on the 5 classes it is already added, with inheritance to VtlScheme.

sosna (approved):
Thanks, @javihern98. Nice work 😊

PR #146, opened 2024-12-06
Make sure all model classes are serializable and deserializable

### LINKED ISSUE #138: Deserialization for item schemes is suboptimal
Item schemes (like Codelist) inherit from ItemScheme. In ItemScheme, the items property has type Item. However, for a Codelist, this type should be Code, for a ConceptScheme, it should be Concept, etc.

This means that, when deserializing a serialized payload (like in a caching scenario), the initial type (Code) is lost and we get Item back instead.

### LINKED ISSUE #139: Cannot deserialize StructureMap
There are issues deserializing StructureMap instances.

The error is:

TypeError: If a type union contains multiple Struct types, all Struct types must be tagged (via `tag` or `tag_field` kwarg) - type `typing.Union[pysdmx.model.map.ComponentMap, pysdmx.model.map.DatePatternMap, pysdmx.model.map.FixedValueMap, pysdmx.model.map.ImplicitComponentMap, pysdmx.model.map.MultiComponentMap]` is not supported

The same holds true for DataflowInfo:

E           msgspec.ValidationError: If a type union contains multiple Struct types, all Struct types must be tagged (via `tag` or `tag_field` kwarg) - type `typing.Union[pysdmx.model.code.Codelist, pysdmx.model.code.Hierarchy, NoneType]` is not supported - at `$.components`

### LINKED ISSUE #140: Add decoders
We have added a utility called encoders under pysdmx.model, to help with the serialization of types that are not Python built-ins.

We need to do something similar for the deserialization. For example:

```python
def decoders(type: Type, obj: Any) -> Any:
    if type is Components:
        comps = []
        for item in obj:
            comps.append(msgspec.convert(item, Component))
        return Components(comps)
    elif type is str and obj.startswith("regex:"):
        p = obj.replace("regex:", "")
        return Pattern(p)
    else:
        raise NotImplementedError(f"Objects of type {type} are not supported")
```

### LINKED ISSUE #143: Problem with serialization of StructureMap
These properties in MultiRepresentationMap create problems when serializing the object:
source: Sequence[Union[str, DataType]] = []
target: Sequence[Union[str, DataType]] = []

The error is: TypeError: If a type union contains multiple Struct types, all Struct types must be tagged (via `tag` or `tag_field` kwarg) - type `typing.Union[pysdmx.model.map.ComponentMap, pysdmx.model.map.DatePatternMap, pysdmx.model.map.FixedValueMap, pysdmx.model....

Later on, there is also this error: Type unions containing a custom type may not contain any additional types other than `None` - type `typing.Union[str, re.Pattern[str]]` is not supported

sosna:
Fix #138 
Fix #139 
Fix #140 
Fix #143

### DIFF
--- a/src/pysdmx/model/__base.py
+++ b/src/pysdmx/model/__base.py
@@ -241,6 +241,20 @@ class MaintainableArtefact(
                 "Maintainable artefacts must reference an agency.",
             )
 
+    def short_urn(self) -> str:
+        """Returns the short URN for the artefact.
+
+        A short URN follows the syntax: Type=Agency:Id(Version). For example:
+        Codelist=SDMX:CL_FREQ(1.0)
+
+        Returns:
+            The short URN for the artefact.
+        """
+        agency = (
+            self.agency.id if isinstance(self.agency, Agency) else self.agency
+        )
+        return f"{self.__class__.__name__}={agency}:{self.id}({self.version})"
+
 
 class ItemScheme(MaintainableArtefact, frozen=True, omit_defaults=True):
     """ItemScheme class.
--- a/src/pysdmx/model/__base.py
+++ b/src/pysdmx/model/__base.py
@@ -278,6 +278,10 @@ class DataflowRef(Struct, frozen=True, omit_defaults=True):
         id: The dataflow identifier (e.g. BIS_MACRO).
         agency: The organisation (or unit) responsible for the dataflow.
         version: The version of the dataflow (e.g. 1.0).
+        name: The name of the dataflow. This is optional as, typically,
+            a dataflow reference wouldn't have this information, but it
+            has been added as it can be useful in a data discovery
+            scenario, for example, in category scheme queries.
     """
 
     agency: str
--- a/src/pysdmx/model/__base.py
+++ b/src/pysdmx/model/__base.py
@@ -265,6 +265,7 @@ class DataflowRef(Struct, frozen=True, omit_defaults=True):
     agency: str
     id: str
     version: str = "1.0"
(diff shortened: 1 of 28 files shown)


src/pysdmx/model/__base.py
```diff
@@ -269,6 +283,7 @@ class DataflowRef(Struct, frozen=True, omit_defaults=True):
     agency: str
     id: str
     version: str = "1.0"
+    name: Optional[str] = None
```
javihern98, line 286: Missing documentation on name attribute, how would it be used?

src/pysdmx/model/__base.py
```diff
@@ -241,6 +241,20 @@ def __post_init__(self) -> None:
                 "Maintainable artefacts must reference an agency.",
             )
 
+    def short_urn(self) -> str:
```
javihern98, line 244: We need to ensure therefore that all model classes have the same name as it is specified in the URN elements. As far as I could check this will work fine.

javihern98 (approved):
All good, just minor things to check again. Approved in any case as it is just to fix documentation and ensure we generate correctly the short_urn on Maintainable.

src/pysdmx/model/__base.py
sosna: Thanks, @javihern98. I checked all class names again and, indeed, they are 100% in line with the SDMX Information Model 😊

src/pysdmx/model/__base.py
sosna: Thanks for spotting this, @javihern98. I have updated documentation accordingly.

PR #147, opened 2024-12-09
Fetch VTL transformation schemes from the FMR

### LINKED ISSUE #92: Add retrieval of VTL artefacts to the Registry clients

sosna:
Fix #92


javihern98 (approved):
All good,, we will modify in a subsequent PR the documentation and checks on some model classes on VTL and add the parsers for SDMX-ML files (#135)

PR #152, opened 2024-12-13
141 migrate sdmx 21 data writers

javihern98:
Summary of changes:

- Implemented the SDMX-ML 2.1 Generic and Structure Specific writers for All-Dimensions and Series
- Implemented a Parser for Short URN for convenience, as we need to extract the individual elements
- Implemented a specific validation before writing for PandasDataset
- Added short_urn property to Schema (for convenience)
- Added attribute dataset_references to Header, to set the dataset references and dimension at observation inside the header when writing SDMX-ML messages
- Adapted SDMX-ML writer to use same function for both Structures writing and Data Writing

All tests for data writers includes a read_xml method to ensure we are writing correctly the data writers, asserting as well the data written shape to ensure we are not missing any datapoint.


src/pysdmx/model/message.py
```diff
@@ -36,7 +36,7 @@ class ActionType(Enum):
     Information = "information"
 
 
-class Header(Struct, frozen=True, kw_only=True):
```
sosna, line 39: When `frozen` is set to `True`, `msgspec` automatically generates a `__hash__` method, which is something that can be useful from time to time. However, it also disables attribute assignment, which can be inconvenient.

If you think the `__hash__` method can be useful for this class but you still want to be able to set attribute values after the object has been created, you can use the [msgspec.structs.replace function](https://jcristharif.com/msgspec/api.html#msgspec.structs.replace).

So, I'm fine with the removal of the `frozen` flag, but I thought I should mention this, just in case you were not aware of this option and find it better.

javihern98: Yes I do need to edit this class with the dataset references, at this moment it is not necessary to have it frozen

src/pysdmx/util/__init__.py
```diff
@@ -83,6 +84,20 @@ def parse_item_urn(urn: str) -> ItemReference:
         raise Invalid(NF, f"{urn} does not match {item_urn_pattern}.")
 
 
+def parse_short_urn(urn: str) -> Reference:
```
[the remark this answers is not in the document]
javihern98: Thanks 😊

src/pysdmx/model/message.py
javihern98: To give more context, the xml writer will only use this attribute (and modify it) to set the structure references for this particular message, i.e. get the datasets we want to write and use their IDs and dimension at observation

PR #157, opened 2024-12-18
Add support for provision agreement queries

### LINKED ISSUE #154: Add support for retrieving provision agreements
We now have a use case where we need to be able to retrieve provision agreements

sosna:
Fix #154


javihern98 (approved):
All good 😊

PR #159, opened 2024-12-18
Add Read_SDMX method

javihern98:
Summary of changes:

- Added convenience method read_sdmx to use same method for all parsers.
- Added SDMXFormat enumeration with the supported formats
- Refactored message.py to avoid circular imports and place there only the Message and Header classes
- Enhanced input_processor to check for xml, csv and json files
- Fixed readers to use always Short_URN as keys of the result, to be placed on Message structure or data fileds
- Changed data type on Component.concept to Union[Concept, ItemReference] to allow the reading of DataStructureDefinition files without references.
- Added support for URL retrieving
- Made httpx dependency as mandatory


src/pysdmx/io/reader.py
```diff
@@ -0,0 +1,112 @@
...
+from pysdmx.io.input_processor import process_string_to_read
+from pysdmx.model.dataset import Dataset
+from pysdmx.model.message import Message
+
+
+class ReadFormat(Enum):
+    """Enumeration of supported SDMX read formats."""
+
+    SDMX_ML_2_1 = "SDMX-ML 2.1"
+    # SDMX_JSON_2 = "SDMX-JSON 2.0.0"
```
javihern98, line 18: @sosna This is a placeholder to include as well emthods to read the JSON formats, still I have not found the way to implement them. Any ideas?

sosna (commented):
Hi @javihern98. There is a lot of work we need to do next year about formats serialization and deserialization. Is there any important reason to "rush" this now, a bit as a last minute addition, or is this something we can discuss when there is more capacity early next year?

javihern98:
> Hi @javihern98. There is a lot of work we need to do next year about formats serialization and deserialization. Is there any important reason to "rush" this now, a bit as a last minute addition, or is this something we can discuss when there is more capacity early next year? 

We do not need to rush it and discuss it further after Christmas😁

Issue #158, opened 2024-12-18
Add seamless integration with vtlengine


javihern98:
- Add vtlengine dependency
- Perform syntax validation on VTL Artefacts (Where necessary)
- Add support for semantic validation and execution from a Dataset

javihern98:
This requires the update of vtlengine and a new release to include the VTL 2.1 standard datastructures: https://github.com/Meaningful-Data/vtlengine/issues/52

javihern98:
Add as well the DataStructureDefinition -> Schema method

javihern98:
Agreed on meeting to postpone this functionality after 1.0

PR #162, opened 2024-12-19
161 pysdmx to vtlengine

mla2001:


javihern98:
Dio not merge, duplicated with #158

PR #167, opened 2025-01-07
Allow both DataflowRef and Dataflow in Category

sosna:
Dataflow was allowed in the CategoryScheme but not the Category. This issue has now been addressed.

Also, I have added a badge to awesome official statistical software.


javihern98 (approved):
All good 😊

PR #170, opened 2025-01-10
164 checking for inconsistencies in parsers and writers methods

mla2001:
Updated XML reader and writers with the new signature.

**reader:**
`def read(infile: Union[str, Path], validate: bool = False, use_dataset_id: bool = False)`
**writer:**
`def write(datasets: Sequence[PandasDataset], output_path: str = "", prettyprint: bool = True, header: Optional[Header] = Header(), dimension_at_observation: Optional[Dict[str, str]] = None)`


src/pysdmx/io/xml/sdmx21/writer/structure.py
```diff
@@ -611,3 +641,53 @@ def write_structures(content: Dict[str, Any], prettyprint: bool) -> str:
     outfile = outfile.replace("& ", "&amp; ")
 
     return outfile
+
+
+def write(
+    datasets: Sequence[STR_TYPES],
```
javihern98, line 647: Change to structures, these are not datasets when we write metadata

src/pysdmx/io/xml/sdmx21/writer/structure.py
```diff
@@ -611,3 +641,53 @@ def write_structures(content: Dict[str, Any], prettyprint: bool) -> str:
...
+
+    outfile += get_end_message(type_, prettyprint)
+
+    if output_path == "":
+        return outfile
+
+    with open(output_path, "w", encoding="UTF-8", errors="replace") as f:
+        f.write(outfile)
+
+    return None
```
javihern98, line 693: This is redundant, please remove it

src/pysdmx/io/xml/sdmx21/writer/structure.py
mla2001: Renamed

src/pysdmx/io/xml/sdmx21/writer/structure.py
mla2001: It raises a mypy error if the return None is missing

PR #172, opened 2025-01-13
IO Package refactor

### LINKED ISSUE #156: Add read_sdmx convenience method
Add the read_sdmx method:

- Supports all readers implemented in io package
- Uses the input processor to use a file path or string buffer or URL
- Infers the format to use from an enumeration that holds the supported formats

### LINKED ISSUE #164: Checking for inconsistencies in parsers and writers methods
Check all parsers and writers follow these guidelines:

- Same arguments order: data, path, (modifiers, mapping, etc)
- Same name for same functions: 
  - Public method names must be "read" and "write"
  - Private and local methods will have _ or __ respectively
- Input data/structures must be a dictionary
  - Typing must be defined as precise as possible, avoiding Any
  - Allow for several datasets writing on each method
  - Add checks to avoid any possible misuse of the methods (example: pass an ItemScheme to a data writer, or pass a Dataset to structures writing)

javihern98:
This PR is a combination of #164 and #156 with many consistency checks. It covers the whole IO package for better consistency

Closes #164
Closes #156 

# Summary of changes

## Dependency management
- Removed extra FMR as we make httpx a mandatory dependency (we need it as well in IO package to download files from URL in read_sdmx)
- Updated Poetry.lock

### Consistency checks (#164)
- Changed signature from readers and writers as specified in #164. Normalization across the whole IO package to use Sequence
- Changed short_urn to be a property on Maintainable Artefact
- Ensure we use Dataflow=... and DataStructure=... across the IO package as short_urn for Schemas
- Removed MessageType and used SDMXFormat on writers
- Updated signature of input processor to use same terminology all across the functions

### Convenience methods (#156)
- Changed input processor to infer SDMX format and flavor (subformat, like SDMX-ML 2.1 Generic)
- Added SDMXFormat enumeration with supported flavors
- Added read_sdmx as a general reader for Path, URL or Strings
- Added get_datasets to extract the Sequence of Datasets with the possibility to use a structures file to extract the Schema

## Model package
- Moved ActionType to Dataset (makes sense only here)
- Cleaned up Message module and used same typing as readers and writers
- Moved SubmissionResult (from RegistryInterface) to submission.py
- Tag on Concept and Item Reference (Util package) due to the use of Concept Identity of both of them in the Component class

### Tests changes
- Adapted tests to the changes above and removed extra tests. Code Coverage reached at 100%


src/pysdmx/io/enums.py
```diff
@@ -0,0 +1,21 @@
...
+
+class SDMXFormat(Enum):
+    """Enumeration of supported SDMX read formats."""
+
+    SDMX_ML_2_1_STRUCTURE = "SDMX-ML 2.1 Structure"
+    SDMX_ML_2_1_DATA_STRUCTURE_SPECIFIC = "SDMX-ML 2.1 StructureSpecific"
+    SDMX_ML_2_1_DATA_GENERIC = "SDMX-ML 2.1 Generic"
+    SDMX_ML_2_1_REGISTRY_INTERFACE = "SDMX-ML 2.1 Registry Interface"
+    SDMX_ML_2_1_ERROR = "SDMX-ML 2.1 Error"
+    SDMX_JSON_2 = "SDMX-JSON 2.0.0"
```
sosna, line 14: I think we should also distinguish between Structure and Data, as we do with XML?

src/pysdmx/io/enums.py
```diff
@@ -0,0 +1,21 @@
+"""IO Enumerations for SDMX files."""
+
+from enum import Enum
+
+
+class SDMXFormat(Enum):
```
sosna, line 6: There is a bit of an overlap with the `Format` class defined in the `pysdmx.api.fmr` module. Should we harmonize them? If so, I would suggest we call the Enum `Format` (after all, Fusion-JSON is not an officially SDMX format. Happy to discuss. Of course, their purpose is a bit different (and I need the media-type for the REST queries), but maybe it's worth considering anyway?

src/pysdmx/model/dataflow.py
```diff
@@ -439,6 +440,23 @@ class DataStructureDefinition(MaintainableArtefact, frozen=True, kw_only=True):
 
     components: Components
 
+    def to_schema(self) -> Schema:
```
sosna, line 443: Why is this necessary? The reason `Schema` was introduced was to represent the response of an SDMX-REST Schema query. In this context, it contains very useful pieces of information like the `artefacts` property, which should list the URNs of all the artefacts that have been considered to generate the schema. If you don't provide this piece of information, then I don't see what converting to a Schema brings as benefits... Just a question :)

src/pysdmx/model/submission.py
```diff
@@ -0,0 +1,20 @@
+"""A module to represent a Submission Result."""
+
+from msgspec import Struct
+
+
+class SubmissionResult(Struct, frozen=True):
```
sosna, line 6: Would it make sense to move this in the message.py file, as it's kind of the result of submitting a message?

src/pysdmx/io/csv/sdmx10/writer/__init__.py
```diff
@@ -1,35 +1,42 @@
...
 from copy import copy
-from typing import Optional
+from typing import Optional, Sequence
 
 import pandas as pd
 
 from pysdmx.io.pd import PandasDataset
 
 
 def writer(
```
sosna, line 11: Why is the function called `writer` for csv and `write` for xml?

src/pysdmx/io/enums.py
javihern98: Sure, let me check the Standard, in any case this has no impact in the rest of the code as we have a NotImplemented

src/pysdmx/io/enums.py
javihern98: Sure, shall we place then this enums at the pysdmx folder? (Root of the source code)

src/pysdmx/io/csv/sdmx10/writer/__init__.py
javihern98: Its a typo, let me change it

src/pysdmx/model/submission.py
javihern98: Maybe, but I do prefer to leave the module message.py as clean as possible

src/pysdmx/io/csv/sdmx10/writer/__init__.py
sosna: Great, thanks for the confirmation 😊. Just in case: The documentation (tutorial) needs to be updated too.

src/pysdmx/io/enums.py
sosna: For me, having the enum about formats in the `io` module looks good, i.e. no need to change the location.

src/pysdmx/model/dataflow.py
sosna: Can we discuss this? Thanks :)

src/pysdmx/io/csv/sdmx10/writer/__init__.py
javihern98: Yes, I will make a general review after merging this, to ensure I can test properly the code snippets

src/pysdmx/io/enums.py
javihern98: Agreed to differentiate between SDMX-JSON_2_DATA and SDMX-JSON_2_STRUCTURE

src/pysdmx/io/enums.py
javihern98: Will be addressed the name change in another issue as we need to harmonize the api.fmr.Format and the io.enums.SDMXFormat into one, using as key the granular format (SDMX-XX VERSION MESSAGE_TYPE) and as value the MIME type

src/pysdmx/model/dataflow.py
javihern98: After discussion:

- I will add the artefacts when adding the Schema. This includes the Concepts, Codelists and DataConstraints (not implemented as of today) of the DataStructureDefinition
- We need to close the gap as much as we can between the Schema and DataStructureDefinition, Dataflow and ProvisionAgreement to make these classes interchangeable. This will be addressed in future releases

src/pysdmx/io/csv/sdmx10/writer/__init__.py
javihern98: No further action is needed here

src/pysdmx/model/dataflow.py
javihern98: Artefacts generated on latest changes

src/pysdmx/io/enums.py
javihern98: Will be addressed when using the general Format enumeration from #175

sosna: Thanks, @javihern98. Unfortunately, there is no official media type for Registration queries. We have opened a new [issue for the SDMX-TWG](https://github.com/sdmx-twg/sdmx-rest/issues/238), but we don't know yet when the media type will be defined. In the meantime, we suggest using the generic `application/xml`.

Regarding the errors, we don't really understand the request. Yes, there were indeed schemas for errors in SDMX-ML 2.1 (i.e. XML only), but they were used only as part of other messages (such as structure messages, generic data messages, etc.). So why is it necessary to list it as "independent" format?

sosna: I have just sent a commit to this branch adding the missing RegistryInterface.

javihern98: Well there is in fact one, at least there is a Simple Error defined in SDMXMessage.xsd:
https://github.com/Meaningful-Data/sdmx-schemas/blob/main/sdmxschemas/xml/sdmx21/SDMXMessage.xsd
(line 1080). I believe this should be as well a different MIME type as it really shows an Error Message and has a different structure

javihern98: I will use in any case application/xml

javihern98: @sosna It seems that using the same value for both keys of the enum (RegisrtyInterface and Error) generates a conflict, so we must use a different MIME for these two. Any ideas? Shall we replicate the structure of the other MIME?

src/pysdmx/io/format.py
```diff
@@ -30,6 +31,7 @@ class Format(Enum):
     STRUCTURE_SDMX_JSON_2_0_0 = f"{_BASE}structure+json;version=2.0.0"
     STRUCTURE_SDMX_ML_2_1 = f"{_BASE}structure+xml;version=2.1"
     STRUCTURE_SDMX_ML_3_0 = f"{_BASE}structure+xml;version=3.0.0"
+    ERROR_SDMX_ML_2_1 = "application/xml"
```
sosna, line 34: I still believe this is unlikely to be necessary, unless you have encountered such messages already? It seems to be more useful in the context of the SOAP web services, which has been deprecated since quite a while? This being said, if you need this, no objection from my side.

src/pysdmx/model/dataflow.py
```diff
@@ -439,6 +440,39 @@ class DataStructureDefinition(MaintainableArtefact, frozen=True, kw_only=True):
...
+            if c.local_codes:
+                out.append(c.local_codes.urn)
+            # Concept URNs
+            if isinstance(c.concept, Concept):
+                out.append(c.concept.urn)
+            else:
+                urn_header = "urn:sdmx:org.sdmx.infomodel.conceptscheme."
+                out.append(urn_header + str(c.concept))
+        result = list({a for a in out if a})
+        return result
```
sosna, line 456: Would this method return the URN of a codelist defined in the concept (i.e. core enumeration)? Also, does it add the URN of the data structure? Ideally, it should also contain the URN of the agency scheme(s) containing the agencies maintaining structures relevant for the data structure.

sosna (commented):
Apparently, the tests create 2 files and don't clean them up after running. The files are called test_output.xml and test_ouput_data.xml.

It could also be due to the fact that there are some failing tests, but it would be great if created files could be removed even if a test fails.

javihern98:
> Apparently, the tests create 2 files and don't clean them up after running. The files are called test_output.xml and test_ouput_data.xml.
> 
> It could also be due to the fact that there are some failing tests, but it would be great if created files could be removed even if a test fails.

Yes, I have added some code to remove these files

src/pysdmx/io/format.py
javihern98: It is in fact the standard error message shown by the BIS API, among other institutions:

https://stats.bis.org/api/v2/schema/datastructure/BIS/BIS_DER_1/1.0

src/pysdmx/model/dataflow.py
javihern98: Updated the code to include Codelist of the Core Representation.

sosna (approved):

PR #173, opened 2025-01-16
Annotations title

### LINKED ISSUE #171: Failure to retrieve schemas
Hello - I am trying to retrieve schemas using the `get_schema` method, but get a validation error. Please see the reproducible code chunk and screenshot of the error below:

```
import pysdmx as px
from pysdmx.api import fmr
imf = fmr.AsyncRegistryClient("https://sdmxcentral.imf.org/sdmx/v2/")
# Checking that the connection works
resp = await imf.get_codes("IMF", "CL_AREA", "1.16")
resp
# Failing request
schema = await imf.get_schema("datastructure", "IMF", "BOP", "2.3")
``` 
image

sosna:
Fix bug affecting annotations in Fusion-JSON. The Fusion-JSON annotation class expected properties like title to be present but these are optional according to the SDMX information model.

Fix #171

### DIFF
--- a/src/pysdmx/io/json/fusion/messages/code.py
+++ b/src/pysdmx/io/json/fusion/messages/code.py
@@ -41,13 +41,16 @@ class FusionCode(Struct, frozen=True):
     def __get_val(
         self, a: FusionAnnotation
     ) -> Tuple[Optional[datetime], Optional[datetime]]:
-        vals = a.title.split("/")
-        if a.title.startswith("/"):
-            return (None, self.__handle_date(vals[1]))
+        if a.title:
+            vals = a.title.split("/")
+            if a.title.startswith("/"):
+                return (None, self.__handle_date(vals[1]))
+            else:
+                valid_from = self.__handle_date(vals[0])
+                valid_to = self.__handle_date(vals[1]) if vals[1] else None
+                return (valid_from, valid_to)
         else:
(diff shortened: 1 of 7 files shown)


src/pysdmx/io/json/fusion/messages/code.py
```diff
@@ -41,8 +41,8 @@ def __handle_date(self, datestr: str) -> datetime:
     def __get_val(
         self, a: FusionAnnotation
     ) -> Tuple[Optional[datetime], Optional[datetime]]:
-        vals = a.title.split("/")
-        if a.title.startswith("/"):
+        vals = a.title.split("/")  # type: ignore[union-attr]
```
javihern98, line 44: Could a.title be None in any case? If so it will raise an unhandled exception here. Please double check this ignore.

javihern98 (approved):
All good, just a minor check is needed. It is approved in any case.

src/pysdmx/io/json/fusion/messages/code.py
sosna: Thanks, @javihern98. No, it could not. The reason is that, before calling the `__get_val` method, we check that the annotation type is set to `FR_VALIDITY_PERIOD`. This is a feature of the FMR, and when type is `FR_VALIDITY_PERIOD`, the `title`property will always be set.

PR #174, opened 2025-01-16
Deser minval

sosna:
The type for minValue, maxValue, startValue and endValue is different between SDMX-JSON and Fusion-JSON (number vs. string). This was causing problems when deserializing objects, when using Fusion-JSON to download the data, as the model class expects an int or a float, but not a string.


javihern98 (approved):
All good😁

PR #195, opened 2025-01-30
Add timeout to RestService

### LINKED ISSUE #193: Allow to pass the desired time out when instantiating a service
It would be nice to set how long we should wait before timing out when executing a rest query.

This should be added to the `pysdmx.api.qb` RestService client.

sosna:
The purpose of this pull request is to add the possibility of setting a timeout when executing a web service query. At the same time, documentation has been improved and a class attribute has been made private.

Closes #193.


javihern98 (approved):
All good 😊

PR #196, opened 2025-01-30
Add asynchronous client for SDMX-REST services

### LINKED ISSUE #194: Add an async service client
The `pysdmx.api.fmr` module offers both a sync and an async clients. The `pysdmx.api.qb` module however only offers a sync client. 

It would be nice to add an async client and then use them in the `pysdmx.api.fmr` module.

sosna:
The purpose of this pull request is to:

- Add an aynchronous client for SDMX-REST services
- Replace the httpx code in the `pysdmx.api.fmr` module with the SDMX-REST synchronous and asynchronous clients

Close #194

### DIFF
--- a/src/pysdmx/api/fmr/__init__.py
+++ b/src/pysdmx/api/fmr/__init__.py
@@ -10,7 +10,6 @@ from typing import (
     Union,
 )
 
-import httpx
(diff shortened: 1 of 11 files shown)


javihern98 (approved):
All good😊

PR #199, opened 2025-01-31
Handle empty item schemes in SDMX-JSON and Fusion-JSON

### LINKED ISSUE #191: The fmr module does not handle most item schemes well when they are empty
The `fmr` module throws an error when retrieving empty item schemes of certain types.

This does not affect codelists, but concept schemes and agency schemes are affected. Other types of item schemes (data provider schemes, category schemes) or collections (e.g. hierarchies) may be affected too.

sosna:
The SDMX-JSON and Fusion-JSON deserializers could not handle empty item schemes for certain types of item schemes. The affected types were: category schemes, concept schemes, agency schemes and data provider schemes. Codelists, value lists and hierarchies were not affected.

Close #191


javihern98 (approved):
All good😊

PR #203, opened 2025-02-06
Wrong media type for mapping queries in the fmr module (SDMX-JSON only)

### LINKED ISSUE #201: Issue with Accept header set for mapping queries in the fmr module
The `fmr` module is expected to set the accept header to `application/vnd.sdmx.metadata+json;version=2.0.0` in case:

- The query is a query for reference metadata (`is_ref_meta` is set to `True`)
- And the requested format is `StructureFormat.SDMX_JSON_2_0_0`

This is the code:

```python
if (
    is_ref_meta
    and self.format == StructureFormat.SDMX_JSON_2_0_0
):
    h = self.headers.copy()
    h["Accept"] = (
        "application/vnd.sdmx.metadata+json;version=2.0.0"
    )
```

Unfortunately, `is_ref_meta` is also set to `True` when querying for representation maps and structure maps, in version 1.0.0 of pysdmx.

The issue has already been fixed thanks to the changes introduced in PR #196 (part of the 1.1.0 release). However, we should still add a test reproducing the issue, to prevent regressions in the future.

sosna:
Close #201 

The issue highlighted by the issue #201 was fixed by PR #196. However, the extra tests aim at preventing regressions later.

### DIFF
--- a/tests/api/fmr/agency_checks.py
+++ b/tests/api/fmr/agency_checks.py
@@ -2,10 +2,13 @@ import httpx
 import pytest
 
 from pysdmx.api.fmr import AsyncRegistryClient, RegistryClient
+from pysdmx.io.format import Format
 from pysdmx.model import Organisation
 
 
-def check_orgs(mock, fmr: RegistryClient, query, body):
+def check_orgs(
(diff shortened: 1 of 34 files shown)


javihern98 (approved):
All good, thanks! 😊

PR #204, opened 2025-02-06
Prepare release of v1.1.0

sosna:
This PR:

- Bumps the version number
- Adds a release note for 1.1.0
- Sets the development status to Production/Stable

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,9 +1,9 @@
 Changelog
 =========
 
-All notable changes to this project are listed in the *release notes* published
-with the `project documentation <https://bis-med-it.github.io/pysdmx>`_ .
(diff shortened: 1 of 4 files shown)


sosna:
I was tempted to leave this for the GH release, to avoid the extra work. Maybe we should avoid this page entirely and simply link to the GitHub releases page?

javihern98:
> I was tempted to leave this for the GH release, to avoid the extra work. Maybe we should avoid this page entirely and simply link to the GitHub releases page?

It is always best to have a "single source of truth", so I believe it would be best to use only the GH release page 😊

javihern98 (approved):
All good, thanks!😊 A new milestone achieved!

PR #212, opened 2025-02-07
Fixed bug on writing SDMX ML file with optional attributes

### LINKED ISSUE #209: Writing SDMX ML file with optional attributes
Hi,
Let's consider a PandasDataset with a structure containing optional attributes.
If I feed this dataset with a pandas' dataframe which does not contain all the columns, then I am not able to write the dataset into a fie.

I receive KeyErrors during validation.

As a workaround, I had to write a small function which adds the attributes:

def _add_missing_fields(df: pd.DataFrame, attributes: list[str]) -> None:
    for att in attributes:
        if att not in df.columns:
            df[att] = None

In the output file, the attributes do not appear.
Can this be simplified, or is it the expected behaviour?

Regards,
Elrick

javihern98:
Closes #209 

Added a check to ensure the optional attributes are present inside the data, if not, the attributes should not be added to series_codes and obs_codes.

The two test files are just format changes (from another branch maybe?)

### DIFF
--- a/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
+++ b/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
@@ -271,6 +271,7 @@ def get_codes(
     """This function divides the components in Series and Obs."""
     series_codes = []
     obs_codes = [dimension_code, dataset.structure.components.measures[0].id]
+    cols = dataset.data.columns
 
     # Getting the series and obs codes
     for dim in dataset.structure.components.dimensions:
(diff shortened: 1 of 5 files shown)


sosna (commented):
I don't see a new test, reproducing the issue. If it is indeed missing, could we please add one, to avoid regressions later?

What do you think?

javihern98:
> I don't see a new test, reproducing the issue. If it is indeed missing, could we please add one, to avoid regressions later?
> 
> What do you think?

Yes let me add it

sosna (approved):

Issue #221, opened 2025-02-15
Add Header information to Message class


javihern98:
# Overview

Currently the Message class does not hold the Header information, which always gets lost when using any reader (read_sdmx or any SDMX-ML individual reader).

## Tasks to perform
- Add an optional attribute called header to the Message class
```python
header: Optional[Header]
```
- Modify current tests on read_sdmx to check the Header information is properly stored

PR #222, opened 2025-02-15
Implemented dataset action and dataset id on header. Added action attribute to dataset on SDMX-ML data writers

### LINKED ISSUE #211: Handle DataSetID and DataSetAction when writing SDMX messages
The Header struct has got a dataset_action field, but this field is not used when the dataset is written to a SDMX file. I checked the code, and this field is actually not used anywhere.
In addition, the Header does not contain the dataset_id field, this one could be useful.

```
class Header(Struct, kw_only=True):
    """Header for the SDMX messages."""

    id: str = str(uuid.uuid4())
    test: bool = True
    prepared: datetime = datetime.now(timezone.utc)
    sender: str = "ZZZ"
    receiver: Optional[str] = None
    source: Optional[str] = None
    dataset_action: Optional[ActionType] = None
    structure: Optional[Dict[str, str]] = None
```

Here is an extract of the xsd: https://github.com/sdmx-twg/sdmx-ml/blob/master/schemas/SDMXMessage.xsd
```
<xs:element name="DataSetAction" type="common:ActionType" minOccurs="0">
	<xs:annotation>
		<xs:documentation>DataSetAction code provides a code for determining whether the enclosed message is an Update or Delete message (not to be used with the UtilityData message).</xs:documentation>
	</xs:annotation>
</xs:element>
<xs:element name="DataSetID" type="common:IDType" minOccurs="0" maxOccurs="unbounded">
	<xs:annotation>
		<xs:documentation>DataSetID provides an identifier for a contained data set.</xs:documentation>
	</xs:annotation>
</xs:element>	
```

Moreover, the set_id and  action fields of the Dataset class are not reflected in the SDMX file at the dataset level. Interestingly, the action field is set to "Information" by default in the class, but "Replace" appears in the file, see below an example of generated file:

`<mes:DataSet ss:structureRef="NA_SEC" xsi:type="ns1:DataSetType" ss:dataScope="DataStructure" action="Replace">`

javihern98:
Close #211 

- Added dataset action and dataset id to SDMX-ML data writers
- Added better handling of dataset qualifiers on SDMX-ML structure specific writer
- Prevent the use of dataset id and dataset action in SDMX-ML Structures writer

### DIFF
--- a/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
+++ b/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
@@ -239,7 +239,9 @@ def __write_header(
     if header.structure is not None:
         for short_urn, dim_at_obs in header.structure.items():
             references_str += __reference(short_urn, dim_at_obs)
-    if not data_message and (header.dataset_id or header.dataset_action):
+    if not data_message and (
+        header.dataset_id or header.dataset_action or header.structure
+    ):
         raise Invalid(
             "Header must not contain DataSetID or DataSetAction "
             "when writing a Structures Message."
--- a/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
+++ b/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
@@ -143,7 +143,10 @@ def add_indent(indent: str) -> str:
 
 
 def __write_header(
-    header: Header, prettyprint: bool, add_namespace_structure: bool = False
+    header: Header,
+    prettyprint: bool,
+    add_namespace_structure: bool = False,
+    data_message: bool = True,
 ) -> str:
     """Writes the Header part of the message.
 
@@ -151,6 +154,7 @@ def __write_header(
         header: The Header to be written
         prettyprint: Prettyprint or not
         add_namespace_structure: Add the namespace for the structure
+        data_message: If the message is a data message
 
     Returns:
         The XML string
@@ -236,6 +240,11 @@ def __write_header(
     if header.structure is not None:
         for short_urn, dim_at_obs in header.structure.items():
             references_str += __reference(short_urn, dim_at_obs)
+    if not data_message and (header.dataset_id or header.dataset_action):
+        raise Invalid(
+            "Header must not contain DataSetID or DataSetAction "
+            "when writing a Structures Message."
+        )
     return (
         f"{nl}{child1}<{ABBR_MSG}:Header>"
         f"{__value('ID', header.id)}"
--- a/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
+++ b/src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
@@ -230,6 +230,9 @@ def __write_header(
     prepared = header.prepared.strftime("%Y-%m-%dT%H:%M:%S")
     test = str(header.test).lower()
     references_str = ""
+    action_value = (
+        header.dataset_action.value if header.dataset_action else None
+    )
     if header.structure is not None:
         for short_urn, dim_at_obs in header.structure.items():
             references_str += __reference(short_urn, dim_at_obs)
@@ -242,6 +245,8 @@ def __write_header(
         f"{__item('Receiver', header.receiver)}"
         f"{__value('Source', header.source)}"
         f"{references_str}"
+        f"{__value('DataSetAction', action_value)}"
+        f"{__value('DataSetID', header.dataset_id)}"
         f"{nl}{child1}</{ABBR_MSG}:Header>"
     ).replace("'", '"')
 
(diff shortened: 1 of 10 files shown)


tests/io/xml/sdmx21/writer/test_structures_writing.py
```diff
@@ -461,3 +463,14 @@ def test_group_deletion(groups_sample, header):
...
     assert any("BIS:BIS_DER(1.0)" in e.short_urn for e in read_result)
+
+
+def test_invalid_header_for_data(header):
+    header_da = copy.deepcopy(header)
+    header_did = copy.deepcopy(header)
+    header_da.dataset_action = ActionType.Append
+    with pytest.raises(Invalid):
+        write([], header=header_da)
+    header_did.dataset_id = "ID"
```
sosna, line 474: Why is this considered invalid? According to the SDMX-ML 2.1 schema, `DataSetID` is of type `common:IDType`. Is this because this is a test for writing structures? In that case, I would rename the test `test_invalid_structure_header`.

tests/io/xml/sdmx21/writer/test_data_writing.py
```diff
@@ -282,3 +283,27 @@ def test_invalid_dimension_key(content):
...
+    assert "DataSetID>TEST_ID</" in result_spe
+    assert "DataSetID>TEST_ID</" in result_gen
+
+    # Read the result to check for formal errors
+    data_spe = read_sdmx(result_spe, validate=True)
+    data_gen = read_sdmx(result_gen, validate=True)
+    assert data_spe.data[0].action == ActionType.Append
+    # We cannot define action for a specific dataset in a generic message
+    # as the attribute is not defined in the XSD
+    assert data_gen.data[0].action == ActionType.Information
```
sosna, line 309: I don't understand this test.

We create a header and set the dataset action to Append. We then use the header to write both a generic and a structure specific messages. 

We then read the data and check the dataset action. For generic, we expect Information while for structure specific, we expect Append (i.e. the value we passed). This looks both confusing and wrong.

There is a comment indicating that we cannot define action for a specific dataset in a generic message as the attribute is not defined in the XSD. Even if that is true (I didn't check), the header for generic data message still [defines an action](https://github.com/sdmx-twg/sdmx-ml/blob/bc28a2255966a3b8d33ec8f61e19f42e6e193284/schemas/SDMXMessage.xsd#L1218), and this should then apply to all datasets contained in the message.

tests/io/xml/sdmx21/writer/test_structures_writing.py
javihern98: Yes, this is only to write structures, I will rename the test

javihern98: Renamed the test to test_invalid_structure_header

tests/io/xml/sdmx21/writer/test_data_writing.py
javihern98: Need to better discuss this in a call. The problem with DatasetAction is that it is unbounded, so there is no way to know which action belongs to which dataset. If there is only one DatasetAction defined in the Header on an SDMX-ML Generic Data Message I agree it should be part of all datasets. This could be part of this issue and checked as well at issue #221 

In SDMX-ML Structure Specific the main issue is that there may be inconsistencies between the DatasetAction defined in the Header and the qualifier "action" defined on each dataset. We need to find a criteria on which Action we should use, I believe this is not covered by the XSD validation

javihern98: Agreed on the meeting to fix the SDMX-.ML Generic writer to add the action at the Dataset level

javihern98: Changed test to ensure we are prioritizing dataset action over header:DataSetAction.

src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
```diff
@@ -229,9 +233,17 @@ def __reference(urn_structure: str, dimension: str) -> str:
     prepared = header.prepared.strftime("%Y-%m-%dT%H:%M:%S")
     test = str(header.test).lower()
     references_str = ""
+    action_value = (
+        header.dataset_action.value if header.dataset_action else None
+    )
     if header.structure is not None:
         for short_urn, dim_at_obs in header.structure.items():
             references_str += __reference(short_urn, dim_at_obs)
+    if not data_message and (header.dataset_id or header.dataset_action):
```
sosna, line 242: Should we add a check for header.structure? I believe it is not allowed in Structure messages ([cf. schema](https://github.com/sdmx-twg/sdmx-ml/blob/bc28a2255966a3b8d33ec8f61e19f42e6e193284/schemas/SDMXMessage.xsd#L1184C2-L1201C19))?

sosna (approved):
Thanks, Javier, nice work. I have one last question but I have already approved the PR 😊

src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
javihern98: Yes, we need to add it, let me check it

javihern98: Added the check to structures, merging it as it is already approved.

PR #226, opened 2025-02-18
Handle multiple agency and data provider schemes

### LINKED ISSUE #224: Organisation queries only return organisations from one scheme
The JSON formats in the `io` module will only deserialize the payload of one scheme (data provider scheme or agency scheme), even if several are returned. This issue in visible [here](https://github.com/bis-med-it/pysdmx/blob/1ef603252e4136437ced2a73e4f90e049b9f8579/src/pysdmx/io/json/sdmxjson2/messages/org.py#L108C9-L111C10), for example.

This works well to support the use cases of the `fmr` module, but it will not work for generic SDMX-REST queries, which could target more than one scheme.

We should address this issue without breaking the backward compatibility of the `fmr` module.

sosna:
This PR adds support for handling JSON responses with multiple agency or data provider schemes. 

The support has been added in a backward-compatible way, i.e. the `fmr` module is not affected.

In addition, utility methods have been added to the organisation schemes (e.g. `__len__` etc), thereby aligning them with the other item schemes.

Close #224

### DIFF
--- a/src/pysdmx/api/fmr/__init__.py
+++ b/src/pysdmx/api/fmr/__init__.py
@@ -379,10 +379,7 @@ class RegistryClient(__BaseRegistryClient):
         query = super()._providers_q(agency, with_flows)
         out = self.__fetch(query)
         schemes = super()._out(out, self.deser.providers)
-        if schemes:
-            return schemes[0].items
-        else:
-            return ()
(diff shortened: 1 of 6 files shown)


javihern98 (approved):
All good, thanks! 😊

PR #233, opened 2025-02-20
Bump version to 1.1.1

sosna:
The purpose of this PR is to set the version to 1.1.1, so that we can release a new version of pysdmx.


javihern98 (approved):
All good, first patch release! 😊

PR #241, opened 2025-03-18
Add Header information to Message class

jfrancma:
- Add Header.py with methods to extract header from the structure
- Add Header to Message
- Fix tests to match Message with Header


javihern98 (commented):
All good 😊, ready for BIS review

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Args:
+        header: Dictionary with the header information.
+
+    Returns:
+        Header: The header of the SDMX message.
+
+    """
+    parsed_header = Header(
+        id=header.get(HEADER_ID, ""),
```
sosna, line 34: In this code, mandatory elements (ID, Test, Prepared and Sender) set a default value for the get function (e.g. `header.get(HEADER_ID, "")`. If my understanding is correct, this should not be necessary, as these fields are mandatory (unless of course we want to be able to handle invalid files).

At the same time, for optional elements(e.g. Receiver, Source, etc.), we don't set a default value (e.g.  `header.get(RECEIVER)`), but in this case I guess we should as these elements are optional?

Could it be that we should do the opposite, or is my understanding wrong?

Also, in general, should we not use None for optional values, instead of an empty string?

Thanks.

src/pysdmx/model/message.py
```diff
@@ -32,9 +32,9 @@ class Header(Struct, kw_only=True):
     """Header for the SDMX messages."""
 
     id: str = str(uuid.uuid4())
-    test: bool = True
+    test: bool = False
     prepared: datetime = datetime.now(timezone.utc)
-    sender: str = "ZZZ"
+    sender: str = "unknown"
```
sosna, line 37: Normally, Sender must be set. If we want to be able to read invalid messages (i.e. messages without Sender), then we could make the field optional. If we don't want to do that, I'd use ZZZ, as this is the code for when the organisation is unspecified, as can be seen in the CL_ORGANISATION codelist published in the SDMX Global Registry.

javihern98: @jfrancma Please change the Sender default value to ZZZ

src/pysdmx/io/xml/sdmx21/reader/header.py
javihern98: Yes, we need to make changes here.

I believe the id could be "" to prevent any errors, but if the rest of the attributes are not found, then we should not pass them to the Header constructor.

Do you agree? @jfrancma @sosna

jfrancma: - Changed the required fields to only get the values and not to set "" string if they are not in header
- Changed the optional fields to only get the values and set to the returned Header if they are in the readed header

src/pysdmx/model/message.py
jfrancma: Changed to "ZZZ"

src/pysdmx/io/xml/sdmx21/reader/header.py
sosna: Sorry for the belated replies. So, in essence, I guess that we can do the same for both mandatory fields and optional fields, i.e. header.get(FIELD_NAME)?

- For the mandatory fields, we can expect the value to be there;
- For the optional fields, if the field is not there, the get method will return None, which I believe is what we'd want?

Thanks!

jfrancma: As you said @sosna, with both mandatory and not mandatory fields we can use .get("xxx"), mandatory fields are always expected and not mandatory fields will return none if not in header

javihern98: I believe the best way would be to not send to the Header constructor any attribute that we don´t have, to use the default values defined at the Header class

sosna: That sounds good to me, thanks. However, if all fields are optional, then I guess using the header.get would be equivalent. If so, then either ways are fine with me 😊

jfrancma: At this point in the script, all mandatory fields will be there, so there won't be any problemsbut  if the non-mandatory fields aren't there, they won't reach the constructor, so they will remain as none.

In conclusion, we will use heder.get("xxx") to collect all the attributes and those that are not there will not be sent (the mandatory ones will always be there).

All right with that?

sosna: Sounds good to me!

jfrancma: @sosna mypy is giving an error if I leave the required fields as header.get("xxx"), since it interprets that those fields can come as None, although those fields always arrive here.
The solution I propose is to set a value in .get so that it interprets that even if it returns None, another value can be set. In this case, I propose setting the default value that the header itself has by default, i.e., "ZZZ" for sender (header.get("sender", "ZZZ"))

sosna: If mypy comment is not applicable (like in this case where the field may in theory be empty but, in practice, it isn't, as we have validated this before), then mypy can also be silenced by adding a # typing: ignore comment on the offending line.

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -0,0 +1,69 @@
...
+
+    """
+    dict_header = {
+        "id": header.get(HEADER_ID),
+        "test": header.get(TEST),
+        "prepared": header.get(PREPARED),
+        "sender": header.get(SENDER).get("id")  # type: ignore[union-attr]
+        if isinstance(header.get(SENDER), dict)
+        else header.get(SENDER),
+        "receiver": header.get(RECEIVER),
```
sosna, line 40: According to the SDMX-ML schemas, both `Sender` and `Receiver` are of [PartyType](https://github.com/sdmx-twg/sdmx-ml/blob/bc28a2255966a3b8d33ec8f61e19f42e6e193284/schemas/SDMXMessage.xsd#L1376C1-L1397C19). As can be seen there, a `PartyType` **must** have an `ID` and **may** have one or more names, and one or more contact details.

In our pysdmx model, sender and receiver are both optional strings. I assume that we will want these strings to be made of the PartyType id attribute.

If this is correct then, unless I'm mistaken, the code on lines 37-40 might not always work as we want. 

Also, if the above is correct, then I believe some information might be lost (e.g. the names and/or contract details provided). If this is acceptable, then, fine. If not, we might want Sender and Receiver to be optional [organisations](https://github.com/bis-med-it/pysdmx/blob/2d554a363ee9220a364d9a853a23a10ebf6648cb/src/pysdmx/model/__base.py#L176) instead.

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -0,0 +1,69 @@
...
+        "source": header.get(SOURCE),
+        "dataset_action": header.get(DATASET_ACTION),
+        "structure": header.get(STRUCTURE),
```
sosna, line 43: This is an example of a Header Structure element:

```xml
<message:Structure
    structureID="BIS_DER"
    namespace="urn:sdmx:org.sdmx.infomodel.datastructure.DataStructure=BIS:BIS_DER(1.0)"
    dimensionAtObservation="AllDimensions">
    <common:Structure>
        <Ref agencyID="BIS" id="BIS_DER" version="1.0" class="DataStructure"/>
    </common:Structure>
</message:Structure>
```

From what I see, this should fit into our structure property, defined as Optional[Dict[str, str]]. 

Are we confident this is what line 43 does?

src/pysdmx/io/xml/sdmx21/reader/header.py
javihern98: I totally agree with your suggestion @sosna , the main problem is that if a user wants to generate a custom header then they may want to add as well the description of the Organisation object or similar attributes, which is not fully compliant with the SDMX IM.. 

If we use only the ID of the PartyType, for me is more than enough for most cases, but we could as well generate a new object called Party that has the desired attributes (in this case: id, name, contacts). For me, losing the name and contact information is totally acceptable (as of today, I have not seen at all a Sender or Reciever with the name and contact defined in the Header)

src/pysdmx/io/xml/sdmx21/reader/header.py
javihern98: The structure parameter will be only used (and probably updated) in the SDMX-ML writers. I suggest to properly define the structure field and adapt these writers.

The current structure dict has the short_urn as key and value is the Dimension at Observation:

```python
structure = {
    "DataStructure=BIS:BIS_DER(1.0)": "AllDimensions"
```

We may keep it as it is or change it, if needed. The line 43 therefore is incomplete @jfrancma

src/pysdmx/io/xml/sdmx21/reader/header.py
sosna: Thanks, @javihern98!

In general, I think we should try and avoid losing information, while keeping things simple. So, instead of creating a dedicated PartyType, we could still reuse the Organisation class. If, by any chance, someone manually creates a header containing more information (like a description), we can simply ignore (and warn) when writing the message?

jfrancma: Hi @sosna,
 can you send me a example of sender and receiver example to use with more than one name and contact details?
 On the other hand, I don't know what the best option would be to send a warning about the description, since in the middle of an execution I don't think that warning would be very legible.

jfrancma: @sosna Let me know if we need something more in Sender/Receiver

I have this example:
 `<message:Sender id="Unknown">
            <common:Name>name1</common:Name>
            <common:Name>name2</common:Name>
  </message:Sender>`

And the return from the example will be this:
`Sender= {'Name': ['name1', 'name2'], 'id': 'Unknown'}`

src/pysdmx/io/xml/sdmx21/reader/header.py
jfrancma: For structure specific xml I am using namespace to extract the information, but since structure and generic xml do not include namespace, I am using the reference (Ref) to extract that same information and form the string.

In both cases the output string would be the same:

` "DataStructure=BIS:BIS_DER(1.0)": "AllDimensions"`

src/pysdmx/io/xml/sdmx21/reader/header.py
javihern98: Agreed on the meeting:

- There should be only one name
- By default, we should take the English Name
- Use the Organisation as object for the PartyType, add a warning in the writer if the user includes a description or any other attribute that is not present.

jfrancma: Change to only get one name, by default English one.
Added a warning if writing or reading a attibute that is not present.

src/pysdmx/io/xml/sdmx21/__tokens.py
```diff
@@ -33,6 +33,17 @@
...
 DATASET_ID = "DataSetID"
+DATASET_ACTION = "DataSetAction"
+TEST = "Test"
+PREPARED = "Prepared"
+SENDER = "Sender"
+RECEIVER = "Receiver"
+SOURCE = "Source"
+HEADER_ID = "ID"
+NAMESPACE = "namespace"
+DIMENSIONATOBSERVATION = "dimensionAtObservation"
```
sosna, line 44: I think that we should follow naming conventions, as already discussed with Javier. Either we separate words in constant names with an `_` (for example, `DATASET_ACTION`, `HEADER_ID`) or we don't (for example `DIMENSIONATOBSERVATION`). But, whichever of these 2 options we prefer, we then use it consistently.

tests/io/xml/sdmx21/writer/test_header_writing.py
```diff
@@ -0,0 +1,123 @@
...
+from pysdmx.io.xml.sdmx21.writer.__write_aux import (
+    __write_header as write_header_aux,
+)
+from pysdmx.model import Organisation
+from pysdmx.model.message import Header
+
+
+@pytest.fixture
+def header():
+    return Header(
```
sosna, line 14: Could it be that the tests and the model classes are no longer aligned?

I see this test, passing an Organisation as sender (which is fine) but a ... string as receiver. 

Checking the model itself, to try and gain clarity, I see the model still expects strings for both senders and receivers,  which I don't think is what we agreed upon?

```python
class Header(Struct, kw_only=True):
    """Header for the SDMX messages."""

    id: str = str(uuid.uuid4())
    test: bool = False
    prepared: datetime = datetime.now(timezone.utc)
    sender: str = "ZZZ"
    receiver: Optional[str] = None
    source: Optional[str] = None
    dataset_action: Optional[ActionType] = None
    structure: Optional[Dict[str, str]] = None
    dataset_id: Optional[str] = None
```

src/pysdmx/io/xml/sdmx21/__tokens.py
javihern98: @jfrancma Please use `_` on tokens names for DIMENSIONATOBSERVATION -> DIMENSION_AT_OBSERVATION

tests/io/xml/sdmx21/writer/test_header_writing.py
javihern98: Yes, we need to align the typing and check the instantiation tests, @jfrancma please check it

javihern98:
@jfrancma Please update the branch when possible, ideally before next review

tests/io/xml/sdmx21/writer/test_header_writing.py
jfrancma: Changed sender and receiver to be Organisation object and modify tests.

src/pysdmx/io/xml/sdmx21/__tokens.py
jfrancma: Removed DIMENSIONATOBSERVATION token as there was already a token for this one,
 DIM_OBS = "dimensionAtObservation", i didnt see this one before.

tests/io/xml/sdmx21/reader/test_reader.py
```diff
@@ -565,3 +570,52 @@ def test_wrong_flavour_generic(error_str):
...
+def test_message_full(samples_folder):
+    data_path = samples_folder / "message_full.xml"
+    input_str, read_format = process_string_to_read(data_path)
+    assert read_format == Format.DATA_SDMX_ML_2_1_STR
+    result = read_sdmx(input_str, validate=True).header
+
+    assert result.sender.id == "Unknown"
+    assert result.sender.name == "Unknown"
+    assert result.receiver.id == "Not_supplied"
+    assert result.structure == "DataStructure=BIS:BIS_DER(1.0):AllDimensions"
```
sosna, line 592: Here, structure is a string but, according to the model class, we expect a Dict[str,str]

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    dict_header = {
+        "id": header.get(HEADER_ID),
+        "test": header.get(TEST),
+        "prepared": header.get(PREPARED),
+        "sender": __parse_sender_receiver(header.get(SENDER)),  # type: ignore[arg-type]
+        "receiver": __parse_sender_receiver(header.get(RECEIVER)),  # type: ignore[arg-type]
+        "source": header.get(SOURCE),
+        "dataset_action": header.get(DATASET_ACTION),
+        "structure": __parse_structure(header.get(STRUCTURE)),  # type: ignore[arg-type]
```
sosna, line 120: This one is problematic, I think, as the `__parse_structure` returns a string, but, according to the model, it should be a dict. This has been missed by mypy because of the # type: ignore[arg-type] set to the Header.

tests/io/xml/sdmx21/writer/test_header_writing.py
```diff
@@ -0,0 +1,112 @@
...
+        sender=Organisation(
+            id="ZZZ",
+            name="unknown",
+        ),
+        receiver=Organisation(
+            id="unknown",
+        ),
+        source=None,
+        dataset_action=None,
+        structure="DataStructure=MD:TEST(1.0):AllDimensions",
```
sosna, line 27: Same comment about the structure as string not being aligned to the model expecting a dict.

sosna (changes requested):

tests/io/xml/sdmx21/reader/test_reader.py
jfrancma: Change message structure to return Dict[str,str] as expected.

src/pysdmx/io/xml/sdmx21/reader/header.py
jfrancma: Remove type ignore and fix _parse_structure to return dict[str,str].

tests/io/xml/sdmx21/writer/test_header_writing.py
jfrancma: Changed all test to be aligned with structure dict[str,str]

sosna (approved):

PR #254, opened 2025-04-03
Align pyproject.toml with PEP 621

sosna:
This PR aligns the pyproject.toml file to PEP 621. 

From the list of tasks, the following was done:

- Replace tool.poetry section with project
- Set the authors and maintainers with the new formats
- Agree on the specific versions we will support from the dependencies
- Set as extras the specific versions we will support on the optional-dependencies
- For dev and docs, keep the same format

The new file was checked with `poetry check` and the poetry.lock file was regenerated.

The following was **not** done:
- Remove the annotation for Python < 4.0, because this prevents the installation of darglint, a dev dependency.
- Discuss the use of `parsy` as a mandatory dependency, instead of being part of the dc extra. Personally, I'm not sure it should be a mandatory dependency, as it is only used in the dc module.
- Ensure some library dependencies works with the latest version of python3-* APT Packages in Ubuntu 24.04 LTS. I haven't done this, as I don't know which ones aren't working at the moment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,18 +1,17 @@
-[tool.poetry]
+[project]
 name = "pysdmx"
 version = "1.1.1"
 description = "Your opinionated Python SDMX library"
-authors = [
-    "Xavier Sosnovsky <xavier.sosnovsky@bis.org>",
-    "Stratos Nikoloutsos <stratos.nikoloutsos@bis.org>",
-    "Francisco Javier Hernandez del Caño <javier.hernandez@meaningfuldata.eu>",
-    "Mateo de Lorenzo Argelés <mateo.delorenzo@meaningfuldata.eu>"
-]
+license = { text = "Apache-2.0" }
 readme = "README.rst"
-documentation = "https://bis-med-it.github.io/pysdmx"
+requires-python = ">=3.9"
+authors = [
+    { name = "Xavier Sosnovsky", email = "<xavier.sosnovsky@bis.org>" },
+    { name = "Stratos Nikoloutsos", email = "<stratos.nikoloutsos@bis.org>" },
+    { name = "Francisco Javier Hernandez del Caño", email = "<javier.hernandez@meaningfuldata.eu>" },
+    { name = "Mateo de Lorenzo Argelés", email = "<mateo.delorenzo@meaningfuldata.eu>" }
+]
 keywords = ["sdmx", "data discovery", "data retrieval", "metadata", "fmr"]
-repository = "https://github.com/bis-med-it/pysdmx"
-license = "Apache-2.0"
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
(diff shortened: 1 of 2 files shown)


javihern98:
I believe parsy could be an optional dependency only for the dc module

pyproject.toml
```diff
@@ -1,44 +1,48 @@
...
     "Development Status :: 5 - Production/Stable",
     "Intended Audience :: Developers",
     "Intended Audience :: Science/Research",
     "License :: OSI Approved :: Apache Software License",
     "Typing :: Typed"
 ]
+dependencies = [
+    "httpx>=0.*",
+    "msgspec>=0.*",
+    "parsy>=2.1",
```
javihern98, line 25: As stated in the PR description, I am in favor of making parsy an optional dependency only for the dc extra

pyproject.toml
```diff
@@ -1,44 +1,48 @@
...
-pandas = { version = "^2.2.2", optional = true }
-vtlengine = { version = "^1.1rc1", optional = true }
+[project.urls]
+homepage = "https://sdmx.io/tools/pysdmx"
+repository = "https://github.com/bis-med-it/pysdmx"
+documentation = "https://bis-med-it.github.io/pysdmx"
+"Bug Tracker" = "https://bis-med-it.github.io/pysdmx/issues"
+
+[project.optional-dependencies]
+data = ["pandas>=2.2.2"]
```
javihern98, line 35: Pandas version should be >= 2.1.4 to be compliant with Ubuntu 24.04

pyproject.toml
```diff
@@ -1,44 +1,48 @@
...
+dc = ["python-dateutil>=2.9.0.post0"]
```
javihern98, line 36: python-dateutil version should be >= 2.8.2

pyproject.toml
```diff
@@ -1,44 +1,48 @@
...
+vtl = ["vtlengine>=1.1rc1"]
+xml = ["lxml>=5.*", "xmltodict>=0.*", "sdmxschemas>=0.2.0"]
```
javihern98, line 38: lxml should be >= 5.2 and xmltodict should be >= 0.13

pyproject.toml
```diff
@@ -1,44 +1,48 @@
...
+all = ["lxml>=5.*", "xmltodict>=0.*", "sdmxschemas>=0.2.0", "pandas>=2.2.2", "python-dateutil>=2.9.0.post0", "vtlengine>=1.1rc1"]
```
javihern98, line 39: Check these dependencies match as well the above versions

javihern98 (commented):
Overall it is good, just minor comments. You may find the list of Ubuntu24.04 with their versions here:

[libs-ubuntu2404.txt](https://github.com/user-attachments/files/19583769/libs-ubuntu2404.txt)

(I preferred to submit a review instead of just make a comment to address each package one by one in a separate conversation) 😊

pyproject.toml
sosna: This wouldn't be enough, as it is also used in the util module. This in turn is used by `pymedal.model.map.DatePatternMap` and the model classes are part of the core installation.

sosna:
I have sent additional commits.

From the list of tasks, the following was done:

- Replace tool.poetry section with project
- Set the authors and maintainers with the new formats
- Agree on the specific versions we will support from the dependencies
- Set as extras the specific versions we will support on the optional-dependencies
- For dev and docs, keep the same format
- **Ensure some library dependencies works with the latest version of python3-* APT Packages in Ubuntu 24.04 LTS**

The following is still not done:

- Remove the annotation for Python < 4.0, because this prevents the installation of darglint, a dev dependency.
- Remove parsy as a mandatory dependency. I was wrong and parsy is indeed used as mandatory dependency. Upon further checks, this is because, in addition to the `pysdmx.api.dc` module, it is also used in the `pysdmx.util` module. This in turn is used by `pymedal.model.map.DatePatternMap` and the model classes are part of the core installation.

pyproject.toml
javihern98: Maybe we can refactor it on a separate issue and address this. Do you agree?

javihern98:
> The following is still not done:
> 
>     * Remove the annotation for Python < 4.0, because this prevents the installation of darglint, a dev dependency.

Darglint is a project that is currently in maintenance mode, while several validations are already implemented in ruff. Shall we drop this dependency?

pyproject.toml
sosna: I'm not sure because I don't know what there is to address, in this case.

In my understanding, the situation is the following:

1. When defining a DatePatternMap, SDMX relies on Java conventions to represent date/time formats. 
2. This we need to respect, i.e. we store this value "as is" in the `pattern` attribute of the DateTimePattern class.
3. However, as this is a Python library and we may want to use these metadata to perform, say, a structure mapping, we have also added a `py_pattern` attribute to the same class. Like this, we have not only the original pattern (example `yyyy`), but also its Python equivalent (e.g. `%Y`).
4. To translate between the 2, we use `parsy`, to parse the original pattern string and map it to its Python equivalent. I think this is a good library to solve this string parsing problem, with multiple possible variations (cf. the [implementation](https://github.com/bis-med-it/pysdmx/blob/develop/src/pysdmx/util/_date_pattern_map.py)).

Considering the above and the fact that the mapping classes are part of the model, i.e. the core pysdmx installation, what is it you want to address?

sosna:
> Darglint is a project that is currently in maintenance mode, while several validations are already implemented in ruff. Shall we drop this dependency?

I haven't compared the 2 libraries so I don't know if ruff performs the same checks. But, if you feel the checks performed are equivalent, I'm happy to drop that dependency. Less is more 😊

javihern98:
> > Darglint is a project that is currently in maintenance mode, while several validations are already implemented in ruff. Shall we drop this dependency?
> 
> I haven't compared the 2 libraries so I don't know if ruff performs the same checks. But, if you feel the checks performed are equivalent, I'm happy to drop that dependency. Less is more 😊

We can address it in any case in another issue, if needed:

The ruff issues list tracks the implementation of the darglint rules:
https://github.com/astral-sh/ruff/issues/458

I am happy to drop it, we can in any case leave the current pyproject.toml as the main problem with the < 4.0 was in the requires-python attribute, not for the dev dependencies

javihern98 (approved):
All good now after discussion 😊

pyproject.toml
javihern98: Ok, then let´s keep it. I got the impression that we preferred to

sosna:
Thanks, @javihern98! I have removed `darglint`, as the link you sent shows that ruff already covers some of the checks and the intention on their side is to eventually add the missing ones.

PR #267, opened 2025-04-10
Read/Write Header: Added structure parsing of StructureUsage, ProvisionAgreement and URN

### LINKED ISSUE #265: Parsing of SDMX files broken with release 1.2.0
The following has been reported by a colleague:

The issue also affects ECB and OECD services.

===

If I run this code with pysdmx 1.1.1 everything works:

from pysdmx.io.reader import read_sdmx
from requests import get

response = get(
    "https://ec.europa.eu/eurostat/api/dissemination/sdmx/2.1/data/NAMA_10_GDP/A..B1GQ.LU+BE?startPeriod=2015-04-01"
).text

dataframe = read_sdmx(response)
print(dataframe)

However, if I run the same code with pysdmx==1.2.0 I get an attribute error:

Traceback (most recent call last):
  File "c:\Workspace\medau_on_github\medau\_Scratchpad\temptest.py", line 9, in <module>
    dataframe = read_sdmx(response)
                ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\fl007141\envs\medau\Lib\site-packages\pysdmx\io\reader.py", line 68, in read_sdmx
    header = read_header(input_str, validate=validate)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\fl007141\envs\medau\Lib\site-packages\pysdmx\io\xml\sdmx21\reader\header.py", line 151, in read
    return __parse_header(header)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\fl007141\envs\medau\Lib\site-packages\pysdmx\io\xml\sdmx21\reader\header.py", line 125, in __parse_header
    "structure": __parse_structure(header.get(STRUCTURE)),
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\fl007141\envs\medau\Lib\site-packages\pysdmx\io\xml\sdmx21\reader\header.py", line 96, in __parse_structure
    reference = structure.get(STRUCTURE).get(REF)  # type: ignore[union-attr]
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'

javihern98:
# Summary of changes

- Refactored code with the Schemas information to cover more possible Structure information in Header
- Added StructureUsage and ProvisionAgreement (ProvisionAgrement due to known bug in SDMX-ML 2.1) structure parsing
- Added URN compatibility
- Small changes in writer to ensure we can write StructureUsage and ProvisionAgrement structures relationships
- Included ruff format changes (like at api/qb)

Closes #265


jfrancma:
When making this change, it would also be necessary to modify the header writer, since right now the header structure would not see these changes reflected.

javihern98 (commented):
All good, ready for BIS review

src/pysdmx/io/xml/sdmx21/__tokens.py
```diff
@@ -27,6 +27,7 @@
 STR_REF = "structureRef"
 STRUCTURE = "Structure"
 STR_USAGE = "StructureUsage"
+PROV_AGREMENT = "ProvisionAgrement"
```
sosna, line 30: Did you find this value in an SDMX-ML schema or generated file? 

If yes, we need to investigate further.

If not, then, I'd expect the value to read `ProvisionAgreement` (i.e. double `e`). In which case, I guess you'll want to update the constant name too?

javihern98:
> Did you run the code that triggered the issue (cf. code snippet in #265) and, if so, could you please confirm that it worked nicely?
> 
> It would also be good to try with an ECB and an OECD query, if possible, as our colleague indicated that they had a problem with all 3 services. Thank you.

Yes, I ran the code as provided. I tried as well with these URLs:

ECB: https://data-api.ecb.europa.eu/service/data/EXR/M.USD.EUR.SP00.A
OECD:  https://sdmx.oecd.org/public/rest/data/OECD.SDD.NAD,DSD_NAAG@DF_NAAG_I?format=genericdata

They are working fine, no errors have been raised.

src/pysdmx/io/xml/sdmx21/__tokens.py
javihern98: I found them in the sdmx-schemas github repository (SDMX-ML 2.1 SDMXCommon Schema, updated less than a year ago). This is the one we use for XSD schema validation:

https://github.com/Meaningful-Data/sdmx-schemas/blob/4056c29e8f7961858461b0919683b028ec197ce3/sdmxschemas/xml/sdmx21/SDMXCommon.xsd#L1115

javihern98: I checked it is the same content as the Schemas available to download at the SDMX official webpage, they both appear with just one e and I think this is already a known issue

sosna: You are right, this is the current version of the schema and it is indeed with the typo: https://github.com/sdmx-twg/sdmx-ml/blob/v2.1/schemas/SDMXCommon.xsd

This is unfortunate, as this will then be different for the 3.0 schemas. In any case, nothing we can do on our side, so I'll approve the PR.

sosna (approved):

PR #276, opened 2025-04-24
Fix circular import error

### LINKED ISSUE #273: Circular import error
When importing `parse_urn` (from `pysdmx.util`), there can be a circular import issue:

Image

The solution could be to move `Reference` and `ItemReference` to the `pysdmx.model` package.

sosna:
Fix #273 

@javihern98, I have moved the references to the __base.py file but happy to discuss if this is not to your liking.


javihern98 (approved):
All good, thanks😊

sosna:
Thanks, @javihern98 😊

PR #277, opened 2025-04-24
Prevent error when reading ComponentRole in a SDMX-ML Structure file

### LINKED ISSUE #275: Drop information of ConceptRole when reading a file
After discussing the possible implementation of ConceptRole and the decision to not support it at the moment (see #272), this issue will only focus on ensuring the ConceptRole information will be dropped but the rest of the file will be parsed correctly.

When parsing SDMX-ML Message from https://api.imf.org/external/sdmx/2.1/dataflow/IMF.RES/WEO?references=all&detail=full, the following error appears:

Image

javihern98:
Closes #275 

(I deleted as well an old test file that was not in use)


sosna (approved):
Looks great! Thanks @javihern98 😊

PR #281, opened 2025-04-29
Upgrade h11 to 0.16.0

sosna:
The purpose of this PR is to patch a vulnerability in h11.

There was a PR from Dependabot but, somehow, it had more changes than expected and so I redid the patch myself.


javihern98 (approved):
All good, thanks😊

PR #282, opened 2025-04-29
Add version to metadata reports

### LINKED ISSUE #279: Version is missing from metadata report
Metadata reports are versionable but the model class does not allow setting the version.

sosna:
The purpose of this PR is to add the version attribute to metadata reports. Both the model class and the FMR json deserializers have been updated.

Close #279


javihern98 (approved):
All good, thanks! 😊

PR #286, opened 2025-05-05
Add SDMX-ML Data 3.0.0 reader

jfrancma:
- Changed input_processor to return DATA_SDMX_ML_3_0 format
- Change reader.py to accept DATA_SDMX_ML_3_0 format
- Add parse to 3.0 in parse_xml
- Add structure reference format 3.0 to parsers


tests/io/xml/sdmx21/reader/test_reader.py
```diff
@@ -660,3 +660,38 @@ def test_header_xmlns(samples_folder):
...
+    header = result.header
+    assert header.structure == {
+        "DataStructure=BIS:BIS_CBS(1.0)": "TIME_PERIOD"
+    }
+    assert result.data is not None
+    data = result.data[0].data
+    num_rows = len(data)
+    num_columns = data.shape[1]
+    assert num_rows == 132
+    assert num_columns == 19
```
javihern98, line 697: Something is wrong here. The number of columns I believe it should be 20. Check as well that the AVAILABILITY column is present.

```suggestion
    assert num_columns == 20
    assert "AVAILABILITY" in data.columns
```

javihern98 (commented):
A specific test structure for sdmx-ml 3.0 is missing (move the relevant files to the new folders).

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -96,6 +97,9 @@ def __parse_structure(
     elif STR_USAGE in structure:
         structure_info = structure[STR_USAGE]
         sdmx_type = DFW
+    elif PROV_AGREEMENT in structure:
+        structure_info = structure[PROV_AGREEMENT]
+        sdmx_type = PROV_AGREEMENT
     else:
         structure_info = structure[PROV_AGREMENT]
         sdmx_type = PROV_AGREMENT
```
javihern98, line 105: Need to add a comment here that PROV_AGREMENT is a typo that appears only in SDMX 2.1, later solved in SDMX 3.0

src/pysdmx/io/xml/sdmx21/reader/__parse_xml.py
```diff
@@ -17,12 +17,31 @@
...
 }
+SCHEMA_ROOT_30 = "http://www.sdmx.org/resources/sdmxml/schemas/v3_0/"
+NAMESPACES_30 = {
+    SCHEMA_ROOT_30 + "message": None,
+    SCHEMA_ROOT_30 + "common": None,
+    SCHEMA_ROOT_30 + "structure": None,
+    "http://www.w3.org/2001/XMLSchema-instance": "xsi",
+    "http://www.w3.org/XML/1998/namespace": None,
+    SCHEMA_ROOT_30 + "data/structurespecific": None,
+    SCHEMA_ROOT_30 + "data/generic": None,
```
javihern98, line 34: Remove this, generic format is not allowed in SDMX 3.0

src/pysdmx/io/xml/sdmx21/reader/__data_aux.py
```diff
@@ -56,9 +57,12 @@ def __get_ids_from_structure(element: Dict[str, Any]) -> Any:
         id_ = element[REF][ID]
         version = element[REF][VERSION]
         return agency_id, id_, version
-    else:
+    elif URN in element:
         urn = parse_urn(element[URN])
         return urn.agency, urn.id, urn.version
+    else:
+        urn = parse_urn(element)  # type: ignore[arg-type]
+        return urn.agency, urn.id, urn.version
```
javihern98, line 65: Simplify this code, maybe keep the else statement and add an inline if

```python
urn = parse_urn(element[URN] if URN in element else element)
```

src/pysdmx/io/reader.py
```diff
@@ -16,7 +16,7 @@
 from pysdmx.util import parse_short_urn
 
 
-def read_sdmx(
+def read_sdmx(  # noqa: C901
```
javihern98, line 19: We need to attempt again to remove the C901 annotation, maybe use a match-case statement?

javihern98 (commented):
Need just very small changes

src/pysdmx/io/xml/sdmx21/reader/header.py
jfrancma: Added a comment to clarify provision agrement is a typo in 2.1 schemas

src/pysdmx/io/xml/sdmx21/reader/__data_aux.py
jfrancma: simplify code with `urn = parse_urn(element[URN] if URN in element else element)` .

src/pysdmx/io/reader.py
jfrancma: As match-case is not a correct syntax in python 3.9, it should stay as if-elif statament.

src/pysdmx/io/xml/sdmx21/reader/__data_aux.py
```diff
@@ -57,7 +58,7 @@ def __get_ids_from_structure(element: Dict[str, Any]) -> Any:
         version = element[REF][VERSION]
         return agency_id, id_, version
     else:
-        urn = parse_urn(element[URN])
+        urn = parse_urn(element[URN] if URN in element else element)  # type: ignore[arg-type]  # noqa: SIM401
```
javihern98, line 61: Remove the ignore SIM401 and use element.get(URN, element) instead of the if-then-else.

jfrancma: Returned to if-else statement as agreed with @javihern98.

javihern98 (commented):
Ready for BIS review, we will set it as ready after version 1.2.2

jfrancma:
@sosna Of course, that change is planned, I will do it as soon as possible on this same PR.
I will create a new reader for version 3.0 separate from 2.1 and move the common functions to a common package.

jfrancma:
@sosna It's ready for review, thanks for your patience.

sosna (commented):
Thanks, this looks very nice! I tested it with some official sample files published in the SDMX-ML repo in GitHib and it works nicely.

I have one minor suggestion. SDMX 3.1 has just been released. Unless I'm mistaken, the only significant change is the namespace (where we have `v3_1` instead of `v3_0`). If that is a minor change, maybe we should add support for this?

javihern98:
> I have one minor suggestion. SDMX 3.1 has just been released. Unless I'm mistaken, the only significant change is the namespace (where we have `v3_1` instead of `v3_0`). If that is a minor change, maybe we should add support for this?

@sosna I believe this should be part of a different issue to analyze the Schemas on SDMX-ML 3.1, we could have it for 1.5.0

sosna:
> @sosna I believe this should be part of a different issue to analyze the Schemas on SDMX-ML 3.1 and generate the new file structure, we could have it for 1.5.0

Sure, @javihern98, that sounds good to me. Let me approve the PR then.

sosna (approved):

PR #290, opened 2025-05-08
Wrap the utility functions for parsing URNs

### LINKED ISSUE #288: Wrap parse urn utility functions
There are 3 functions to parse an URN, depending on the URN type (maintainable, item or short). It would be nice to have an additional wrapper function around these, so that you don't know to know the type before calling the appropriate function.

sosna:
In this PR, the following changes have been made to the URN parsing utility functions:

- The original `parse_urn` function has been renamed to `parse_maintainable_urn`.
- A function to parse short URNs of items has been added.
- The new `parse_urn` function is now generic, i.e. it will try all specific functions until a match is found. If no match is found, an `Invalid` error will be returned.

Tests have also been added, to verify that URNs of nested items can be parsed too.

The changes above are deemed backward compatible because the new `parse_urn` function can do anything the old could do, i.e. clients should not be impacted by the change. 

Close #288

### DIFF
--- a/docs/api/helper/urn.rst
+++ b/docs/api/helper/urn.rst
@@ -11,8 +11,9 @@ maintained by ``SDMX`` will have the following unique URN:
 
 ``pysdmx`` offers helper methods to:
 
-- Parse URNs of maintainable artefacts (``parse_urn``).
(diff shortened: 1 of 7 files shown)


javihern98 (approved):
All good, thanks! 😊

PR #291, opened 2025-05-08
Add SDMX-ML Data 3.0.0 writer

jfrancma:
- Add sdmx30 package with structure specific writer for version 3.0.
- Create write_structure_specific_aux for common methods of writer structure specifir for versions 2.1 and 3.0.
- Refactor writer_aux and writer_data_aux for version 3.0 .
- Add parse_source and parse_dataset_action in header.py.


src/pysdmx/io/xml/__write_aux.py
```diff
@@ -55,18 +57,31 @@
...
+    ABBR_STR: f"{BASE_URL_21}/structure",
+    ABBR_SPE: f"{BASE_URL_21}/data/structurespecific",
+}
+
+BASE_URL_30 = "http://www.sdmx.org/resources/sdmxml/schemas/v3_0"
+
+NAMESPACES_30 = {
+    "xsi": "http://www.w3.org/2001/XMLSchema-instance",
+    ABBR_MSG: f"{BASE_URL_30}/message",
+    ABBR_GEN: f"{BASE_URL_30}/data/generic",
```
javihern98, line 76: Remove this, generic format is deprecated in SDMX 3.0

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -116,6 +117,26 @@ def __parse_structure(
...
+    if source is None:
+        return None
+    elif isinstance(source, dict):
+        return source["#text"]
+    else:
+        return source
+
+
+def __parse_dataset_action(
+    dataset_action: Union[Dict[str, Any], None],
```
javihern98, line 131: Typing of dataset_action should be Optional[Dict[str,Any]], to include the use of None

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -116,6 +117,26 @@ def __parse_structure(
...
+) -> Union[ActionType, None]:
```
javihern98, line 132: Same as above

tests/io/xml/sdmx30/writer/test_data_writing.py
```diff
@@ -0,0 +1,238 @@
...
+def test_data_write_data_structure_30_no_header(
+    data_structure,
+):
+    result = write_str_spec(
+        datasets=[data_structure],
+        prettyprint=True,
+    )
+    assert result is not None
+    assert result.count("Header") == 2
+    print(result)
```
javihern98, line 226: I cannot understand what is the purpose of this test. What are we trying to cover here? We need to remove the print at the end

src/pysdmx/io/xml/__write_aux.py
```diff
@@ -235,6 +263,7 @@ def __reference(
     nl: str,
     prettyprint: bool,
     add_namespace_structure: bool,
+    references_30: Optional[bool] = False,
```
javihern98, line 266: Typing here could be simply bool, I cannot see why this attribute could be set to None

src/pysdmx/io/xml/__write_aux.py
```diff
@@ -278,6 +326,7 @@ def __write_header(
     prettyprint: bool,
     add_namespace_structure: bool = False,
     data_message: bool = True,
+    references_30: Optional[bool] = False,
```
javihern98, line 329: Typing here could be simply bool, I cannot see why this attribute could be set to None

src/pysdmx/io/xml/sdmx30/writer/structure_specific.py
```diff
@@ -0,0 +1,104 @@
...
+        content, dimension_at_observation
+    )
+    header.structure = dim_mapping
+    add_namespace_structure = True
+    for i, (short_urn, dimension) in enumerate(header.structure.items()):
+        reference = parse_short_urn(short_urn)
+        if reference.sdmx_type == PROV_AGREEMENT:
+            pre_urn = REGISTRY_LOW
+        else:
+            pre_urn = DSD_LOW
```
javihern98, line 71: Try to simplify this with an inline if

tests/io/xml/sdmx30/writer/samples/data_write_30_data_flow.xml
```diff
@@ -0,0 +1,21 @@
...
+			<com:StructureUsage>
+				urn:sdmx:org.sdmx.infomodel.datastructure.Dataflow=MD:TEST(1.0)
+			</com:StructureUsage>
+		</mes:Structure>
+		<mes:Source>PySDMX</mes:Source>
+	</mes:Header>
+	<mes:DataSet ds_att1="value1" ds_att2="10" ss:structureRef="TEST" xsi:type="ns1:DataSetType" action="Information">
+		<Obs DIM1="1" />
+		<Obs DIM1="2" />
+		<Obs DIM1="3" />
```
javihern98, line 19: I am missing the value of a Measure here, please check and change the input dataset if needed to ensure a Measure is present

tests/io/xml/sdmx30/writer/samples/data_write_30_data_structure.xml
```diff
@@ -0,0 +1,21 @@
...
+			<com:Structure>
+				urn:sdmx:org.sdmx.infomodel.datastructure.DataStructure=MD:TEST(1.0)
+			</com:Structure>
+		</mes:Structure>
+		<mes:Source>PySDMX</mes:Source>
+	</mes:Header>
+	<mes:DataSet ss:structureRef="TEST" xsi:type="ns1:DataSetType" action="Information">
+		<Obs DIM1="1" />
+		<Obs DIM1="2" />
+		<Obs DIM1="3" />
```
javihern98, line 19: Also missing the measure values here

tests/io/xml/sdmx30/writer/samples/data_write_30_prov_agree.xml
```diff
@@ -0,0 +1,21 @@
...
+			<com:ProvisionAgreement>
+				urn:sdmx:org.sdmx.infomodel.registry.ProvisionAgreement=MD:TEST(1.0)
+			</com:ProvisionAgreement>
+		</mes:Structure>
+		<mes:Source>PySDMX</mes:Source>
+	</mes:Header>
+	<mes:DataSet ds_att1="value1" ds_att2="10" ss:structureRef="TEST" xsi:type="ns1:DataSetType" action="Information">
+		<Obs DIM1="1" />
+		<Obs DIM1="2" />
+		<Obs DIM1="3" />
```
javihern98, line 19: Also missing the measure values here

javihern98 (commented):
Small changes are still needed

src/pysdmx/io/xml/sdmx21/reader/header.py
jfrancma: changed to optional instead of union.

src/pysdmx/io/xml/__write_aux.py
jfrancma: Changed it to be just bool.

src/pysdmx/io/xml/__write_aux.py
jfrancma: Changed it to be just bool.

tests/io/xml/sdmx30/writer/samples/data_write_30_data_structure.xml
jfrancma: Added a measure value to the test.

tests/io/xml/sdmx30/writer/samples/data_write_30_prov_agree.xml
jfrancma: Added a measure value to the test.

tests/io/xml/sdmx30/writer/test_data_writing.py
javihern98: Ok, then it is necessary

jfrancma: Added better assertion to the test.

javihern98 (commented):
Ready for BIS review, we will set it as ready after release 1.3.0

sosna (commented):
I would like to review this once the SDMX-ML 3.0 reader code has been merged, if that's OK, as this would help with my testing :-). But I have a quick comment already: SDMX 3.1 has just been released and, unless I'm mistaken, as far as SDMX-ML is concerned, the only important different should be in the namespace definition. Do we want to add support for this already?

sosna (changes requested):
There seems to be an issue somewhere. 

I have downloaded the sample files from the SDMX 2.1 official release (cf. [section 3B](https://sdmx.org/wp-content/uploads/SDMX_2-1_SECTION_3B_SDMX_ML_Schemas_Samples_2020-07.zip)). I used the file named `ecb_exr_ng_ts_gf.xml`.

My idea was, for this test, to:

1. Read the SDMX-ML 2.1 StructureSpecific file.
2. Use the new SDMX-ML 3.0 writer to write the content of the datasets and header read in step 1.
3. Use the new SDMX-ML 3.0 reader to read the file back.
4. Assert that the headers and datasets were equal (i.e. lossless conversion).

Here is the python code used for this test:

```python
from pysdmx.io.xml.sdmx30.writer.structure_specific import write

# Read from file
data_msg1 = read_sdmx("data.xml")

# Extract the necessary information
datasets = data_msg1.data
header = data_msg1.header

# Write to SDMX-ML 3.0
write(datasets, "out.xml", header=header)

# Read the written file
data_msg2 = read_sdmx("out.xml")

# Assert the read/write operations were lossless
assert data_msg2.header == header
assert data_msg2.data == datasets
```

However, it fails with the following error:

```console
raise Invalid("Validation Error", ";\n".join(severe_errors))
pysdmx.errors.Invalid: Validation Error: Element '{http://www.sdmx.org/resources/sdmxml/schemas/v3_0/common}Structure': [facet 'pattern'] The value '
                                urn:sdmx:org.sdmx.infomodel.datastructure.DataStructure=ECB:ECB_EXR_NG(1.0)
                        ' is not accepted by the pattern '.+\.datastructure\.DataStructure=.+'.
```

Looking in the generated XML file, I see that the Structure namespace has 2 equal signs (`urn:sdmx:org.sdmx.infomodel.datastructure.DataStructure==ECB:ECB_EXR_NG(1.0)`) but I'm not sure if that is the reason for the failure.

javihern98:
> > Rechecked the data writer and added extra line of code to remove np.nan on SDMX-ML data writers

> Assuming an SDMX-ML input message with an explicitly missing OBS_VALUE (i.e. `"OBS_VALUE" = "NaN"`) and an OBS_STATUS set to `L`, `M`, `O` or `Q` (i.e. codes representing missing values for different reasons). In that case, writing an SDMX-ML file without OBS_VALUE would sound incorrect to me. @stratosn, what do you think?

Hi @sosna I believe that validation is out of the scope of this issue as this latest check was to prevent only np.nan (i.e. null values) to be written as "nan". The validation you propose should be part of the upcoming validation functions for the library and IMO should not be a validation done in the Data writers

sosna:
> Hi @sosna I believe that validation is out of the scope of this issue as this latest check was to prevent only np.nan (i.e. null values) to be written as "nan". The validation you propose should be part of the upcoming validation functions for the library and IMO should not be a validation done in the Data writers. Null values should be only considered as empty commas

It could be that I completely misunderstood but, is it not so that, with your changes, you will generate an invalid output? My point is that, if something that was provided in input is dropped when writing the output, not only is the transformation no longer lossless but, in this case, it might also be that the writer then output something wrong?

sosna:
> I mean it in the opposite way, without this latest change we would generate an invalid output as this data:

But this is why, in my first comment, I distinguished between explicitly and implicitly missing values. The example you gave shows implicitly missing values and, in that case, I agree with the behaviour. The example you showed above is correct, IMO

Where I'm not sure it's correct, is for **explicitly missing** values. For example, when an OBS_VALUE attribute set to NaN in the input SDMX-ML file.

If we get this in input `<Obs OBS_VALUE="NaN" OBS_STATUS="M"/>`, and we now get this in output `<Obs OBS_STATUS="M"/>`, I believe this is wrong (@stratosn, could you please confirm?).

javihern98:
> > II have checked the writer function keeps the NaN values even with this change, as long as we set properly the na_values in the readers.
> 
> I understand the above as meaning that the transformation is lossless (i.e. if OBS_VALUE="NaN" was set in input then it is also present in output). If so, then, there are no objections on my side 😊. Thanks for the clarification, @javihern98!

That is right, the transformation now is lossless, as expected 😊

sosna:
> @sosna @stratosn We are awaiting your approval on this PR if no further changes are needed 😊

Hi @javihern98. Apologies for the delay, yesterday was a busy day. But I'll start reviewing all the PRs, at 9:30 today

sosna:
> @sosna Yes, we should add them, maybe we can use the general writer issue (once it is decided) and place them there.

Perfect, thanks @javihern98 😊. Happy to help with this, if needed!

PR #296, opened 2025-05-15
Fix header source and dataset action reader and prepared writer

jfrancma:
- Fix dataset action on reader to match action type
- Fix Source to read lang from xml files
- Fix Prepared to write correct format


src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -132,8 +153,8 @@ def __parse_header(header: Dict[str, Any]) -> Header:
         "prepared": header.get(PREPARED),
```
javihern98, line 153: Seems something is missing here as well, prepared should always be a datetime object, we would set the attribute prepared in the header object as string. A try-except would be also advisable to prevent any errors and do not parse it if any exception occurs

src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
```diff
@@ -315,7 +315,9 @@ def __value(element: str, value: Optional[str]) -> str:
     nl = "\n" if prettyprint else ""
     child1 = "\t" if prettyprint else ""
     if isinstance(header.prepared, str):
-        prepared = datetime.strptime(header.prepared, "%Y-%m-%dT%H:%M:%S")
+        prepared = datetime.strptime(
+            header.prepared, "%Y-%m-%dT%H:%M:%S"
+        ).strftime("%Y-%m-%dT%H:%M:%S")
```
javihern98, line 320: header.prepared should never be a str here, it should always be a datetime object

javihern98 (commented):
Review the prepared attribute when reading the header and ensure it is always a datetime object.

src/pysdmx/io/xml/sdmx21/reader/header.py
jfrancma: Added parse_prepared, this method try to parse prepared to a datetime if fails, return datatime.now().

src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
jfrancma: Changed to only be datetime.

javihern98:
@sosna We should think about using python-dateutil to parse the prepared data, if not we could have a lot of errors and make it manually. Currently it is part of the dc extra, we could remove it and make it part of the main library

sosna:
@javihern98: I usually try to keep the amount of dependencies of the base install minimal. But, if you think it would make the code more robust, we can definitely do as you suggested. Will you make the changes now, and I wait for the review?

jfrancma:
@sosna @javihern98 With the solution implemented now, there should be no need to use python-dateutil.

src/pysdmx/io/xml/sdmx21/reader/header.py
```diff
@@ -116,6 +118,31 @@ def __parse_structure(
...
+    """Parses the dataset action of the SDMX message."""
+    if dataset_action is None:
+        return None
+    else:
+        return ActionType(dataset_action)
+
+
+def __parse_prepared(prepared: str) -> datetime:
+    """Parses the prepared date of the SDMX message."""
+    return datetime.fromisoformat(prepared.replace("Z", "+00:00"))
```
sosna, line 143: Could this fail, in case it is another time zone code, such as CET or CEST?

src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
```diff
@@ -316,10 +315,9 @@ def __value(element: str, value: Optional[str]) -> str:
 
     nl = "\n" if prettyprint else ""
     child1 = "\t" if prettyprint else ""
-    if isinstance(header.prepared, str):
-        prepared = datetime.strptime(header.prepared, "%Y-%m-%dT%H:%M:%S")
-    else:
-        prepared = header.prepared.strftime("%Y-%m-%dT%H:%M:%S")
+    prepared = header.prepared.isoformat(timespec="seconds").replace(
+        "+00:00", "Z"
```
sosna, line 319: Is this `replace` necessary, i.e. does the SDMX spec mandates that the prepared timestamp must be in the Zulu time zone?

src/pysdmx/io/xml/sdmx21/reader/header.py
jfrancma: @sosna As ir read in the documentation, timezone can only be write as Z or +/-HH:MM.
Here is the line of the documentation: The lexical representation of a timezone is a string of the form: (('+' | '-') hh ':' mm) | 'Z'.
Also if i try to write is in another format than Z or +/-HH:MM the schema validation fails.

src/pysdmx/io/xml/sdmx21/writer/__write_aux.py
jfrancma: @sosna  The use of zulu time zone is not mandatory but it is recommended.
Examples of expected time formats:
   Zulu time zone : 2025-05-22T10:00:00Z
   Offset from UTC: 2025-05-22T08:00:00+/-02:00
   Local time with no time zone: 2025-05-22T10:00:00

sosna (approved):

PR #297, opened 2025-05-15
Build a Query object out of an SDMX Reference or a URN

### LINKED ISSUE #289: Build a Query object out of an SDMX Reference or a URN
We need to add the possibility to create an SDMX query out of a URN or an Reference object.

sosna:
This PR adds a factory method to `StructureQuery` and `RefMetaByMetadatasetQuery`. The factory methods allow creating an instance of these queries out of a reference object or an SDMX URN.

Close #289

### DIFF
--- a/docs/api/query/registration.rst
+++ b/docs/api/query/registration.rst
@@ -1,0 +1,11 @@
+SDMX-REST Registration Queries
+==============================
+
+.. note::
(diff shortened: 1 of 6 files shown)


javihern98 (approved):
All good, thanks! 😊

PR #308, opened 2025-05-21
Fixed get_datasets schema generation

### LINKED ISSUE #301: get_datasets always returns the Schema pointing to a DataStructureDefinition
# Overview

Whenever a dataset references a Dataflow, the get_datasets function returns the Schema referencing the underlying DataStructureDefinition, but it should reference the Dataflow to keep consistency

## Example

```python
from pysdmx.io import get_datasets

data_path = "str_all_minimal_df.xml"
metadata_path = "metadata_minimal_df.xml"

datasets = get_datasets(data_path, metadata_path)
print(datasets[0].short_urn)
```

### Files:
[example_files.zip](https://github.com/user-attachments/files/20232170/example_files.zip)

### Dataset reference
```xml
        <message:Structure structureID="MD_DF"
                           namespace="urn:sdmx:org.sdmx.infomodel.datastructure.Dataflow=MD:TEST_DF(1.0)"
                           dimensionAtObservation="AllDimensions">
            <common:StructureUsage>
                <Ref agencyID="MD" id="TEST_DF" version="1.0"/>
            </common:StructureUsage>
        </message:Structure>
```

### Result
```
DataStructure=MD:DSD_1(1.0)
```

### Expected Result
```
Dataflow=MD:TEST_DF(1.0)
```

## Tasks to perform
- Add to_schema method to Dataflow, it returns a Schema object only if the structure is a DataStructureDefinition, keeping the short_urn of the Dataflow but the components of the DataStructureDefinition
- Modify get_datasets method to keep consistency with the object the Dataset references

javihern98:
Closes #301 

- Replaced to_schema methods to a custom function to generate a Schema from Dataflow/DataStructureDefinition
- Enhanced tests and error messages


jfrancma (commented):
All good, ready for BIS review.

src/pysdmx/model/dataflow.py
```diff
@@ -446,7 +446,7 @@ class DataStructureDefinition(MaintainableArtefact, frozen=True, kw_only=True):
 
     components: Components
 
-    def __extract_artefacts(self) -> Sequence[str]:
+    def _extract_artefacts(self) -> Sequence[str]:
```
sosna, line 449: I'm not sure it's useful to have this method here now, once we remove to_schema (but please, see previous comment about this).

sosna (changes requested):
Overall, I think it's OK, but is there a way to make this change backward compatible?

src/pysdmx/model/dataflow.py
```diff
@@ -463,21 +463,6 @@ def __extract_artefacts(self) -> Sequence[str]:
         result = list({a for a in out if a})
         return result
 
-    def to_schema(self) -> Schema:
```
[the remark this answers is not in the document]
javihern98: Then we should keep only this to_schema method and adapt the schema_generation function to use it 🥲

src/pysdmx/model/dataflow.py
javihern98: Yes, we should keep it as private if we keep the to_schema method

src/pysdmx/util/_model_utils.py
```diff
@@ -0,0 +1,80 @@
...
+                f"Not found referenced DataStructure {dsd_ref}"
+                f"from Dataflow {dataset_ref}. "
+                f"Please send the structures message using "
+                f"references=children to include the DataStructureDefinition.",
+            ) from None
+        return __generate_schema_from_dsd(
+            dsd=dsd, schema_reference=dataset_ref, context=context
+        )
+    elif context == "provisionagreement":
+        raise NotImplementedError(
```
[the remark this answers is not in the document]
javihern98: We simply do not support the Provision Agreement yet when reading SDMX-ML structures, therefore we should support first the ProvisionAgreement and DataProvider classes and change in the same issue this part of the code as well. If we finish the SDMX-JSON readers we could also make the change there

src/pysdmx/model/dataflow.py
```diff
@@ -463,6 +463,21 @@ def _extract_artefacts(self) -> Sequence[str]:
...
+            context="datastructure",
+            agency=(
+                self.agency.id
+                if isinstance(self.agency, Agency)
+                else self.agency
+            ),
+            id=self.id,
+            components=self.components,
+            version=self.version,
+            artefacts=self._extract_artefacts(),
```
sosna, line 478: Now that the code is back in this class, the extract_artefacts method can be made private again, as it was before

src/pysdmx/util/_model_utils.py
```diff
@@ -64,8 +37,13 @@ def schema_generator(message: Message, dataset_ref: Reference) -> Schema:
                 f"Please send the structures message using "
                 f"references=children to include the DataStructureDefinition.",
             ) from None
-        return __generate_schema_from_dsd(
-            dsd=dsd, schema_reference=dataset_ref, context=context
+        return Schema(
```
sosna, line 40: Should we not get the schema via dsd.to_schema().artefacts instead? Like this, we use the public API instead of knowing too much about the inner working of that class...

javihern98: It is a bit redundant IMO, but I find it cleaner as well to keep the __extract_artefacts function as private

src/pysdmx/model/dataflow.py
javihern98: Yes. setting it to private

src/pysdmx/util/_model_utils.py
javihern98: Changed to dsd.to_schema().artefects as suggested

sosna (approved):

PR #311, opened 2025-05-21
Add a validation to raise a invalid error in write_nameable

jfrancma:
- Add `raise Invalid( "Name is required for NameableArtefact" f" id= {nameable.id}")`  error if name attribute is None for NameableArtefacts


src/pysdmx/io/xml/sdmx21/writer/structure.py
```diff
@@ -229,6 +230,10 @@ def __write_nameable(
                     f"</{ABBR_COM}:{attr}>"
                 )
             ]
+        elif attr == "Name":
+            raise Invalid(
+                "Name is required for NameableArtefact" f" id= {nameable.id}"
```
javihern98, line 235: Maybe we can specify in the error message that we require to have the name to **write** a Nameable artefact, @sosna any thoughts on the error message?

javihern98 (commented):
Ready for BIS review, just need to discuss the actual error message

src/pysdmx/io/xml/sdmx21/writer/structure.py
sosna: @javihern98, considering we are using a writer, the error message is already clear, in my opinion. However, if you prefer to add that this is required when writing, it is also OK for me. In short, both work, as far as I'm concerned 😊

sosna (approved):
Looks good, thanks 😊

src/pysdmx/io/xml/sdmx21/writer/structure.py
jfrancma: @sosna if the error message is ok for you, i will merge it then  😊

sosna: Totally fine, thanks 😊

PR #316, opened 2025-05-27
Add generic deserializers for SDMX-JSON 2.0.0 Structure messages

### LINKED ISSUE #299: Add generic serializers/deserializers for SDMX-JSON 2.0.0
There is a lot of deserialization code for SDMX-JSON 2.0.0 Structure messages already but it was written to support the metadata retrieval features exposed by the FMR client. We now need a generic one, so that any type of SDMX-JSON structure messages can be processed.

sosna:
The purpose of this PR is to add a generic reader for SDMX-JSON 2.0.0 structure messages. We already had various deserializers for SDMX-JSON 2.0.0 but they were limited to the use cases required by the FMR clients.

We still only support a subset of the SDMX information model but the missing items will need to be prioritized separately.

Close #299


src/pysdmx/util/_model_utils.py
```diff
@@ -27,7 +27,7 @@ def schema_generator(message: Message, dataset_ref: Reference) -> Schema:
             raise Invalid(
                 f"Dataflow {dataset_ref} does not have a structure defined.",
             )
-        dsd_ref = parse_urn(dataflow.structure)
+        dsd_ref = parse_urn(dataflow.structure)  # type: ignore[arg-type]
```
javihern98, line 30: We need to make a change here and in SDMX-ML structure reader to accept as well the DataStructureDefinition object, to prevent any error when generating the schema in get_datasets function. We could address this in a separate issue as well.

javihern98 (commented):
Code seems good, but the model changes need to be addressed as well in the SDMX-ML readers (dataflow with DataStructureDefinition as structure and typing on return of reading structures. Moreover, it is also missing some way to call a "read" function to maintain consistency with the rest of readers. The read function should be similar to this one:

```python
def read(
    input_str: str
) -> Sequence[MaintainableArtefact]:
```

Same signature must be used in SDMX-ML Structures reader and writer, as it must be the same as the Message/StructureMessage one

src/pysdmx/util/_model_utils.py
javihern98: @sosna Shall we add a separated issue for this to not block this PR?

PR #322, opened 2025-05-28
Prettyprint argument on generate_vtl_script and VTL Toolkit documentation

javihern98:
# Summary of changes

- Added prettyprint argument on generate_vtl_script
- Improved VTL Toolkit documentation
- Added API Reference for VTL toolkit functions
- Fixed access to VTL toolkit functions and renamed files to prevent same name on module and function
- Improved `vtl` extra error message
- Ensure the `vtl` extra check is applied only when the model_validations or prettyprint are called. For this reason, the model_validations default value for generate_vtl_script is set to False as the default behaviour should produce a result without the vtl extra
- Fix on vtlengine version range on `all` extras


jfrancma (commented):
Looks good, ready fos BIS review. 😄

docs/api/toolkit.rst
```diff
@@ -0,0 +1,15 @@
+.. _toolkit:
+
+Toolkit
+=======
+
+The toolkit module offers a collection of utility functions designed to simplify
```
sosna, line 6: Now that I read the description of the toolkit, where should the recently introduced mapping of SDMX data types to Pandas be placed? Under util, as it is now, or rather under toolkit? It seems to me as if I should move them under toolkit but would like to get your opinion first. Thanks!

sosna (approved):
Looks good, thanks 😊.

I just have a question about whether the code under util.bridges should be moved under api/toolkit instead.

javihern98:
By the description I believe it could be placed under toolkit as well😊

docs/api/toolkit.rst
javihern98: By the description I believe it could be placed under toolkit as well😊

Issue #88, opened 2024-10-03
Missing text format facets and data types


stratosn:
Comparing to the SDMX 3.0 schemas, the following text format facets are missing:
```
interval (xs:decimal)
timeInterval (xs:duration)
isMultiLingual (xs:boolean)
```
compared to:
https://github.com/bis-med-it/pysdmx/blob/5aab14598ff4b5cede2282e1ed9f0a77146fe7d7/src/pysdmx/io/json/fusion/messages/core.py#L32

In the [concept facets](https://github.com/bis-med-it/pysdmx/blob/5aab14598ff4b5cede2282e1ed9f0a77146fe7d7/src/pysdmx/model/concept.py#L72) the above are supported, except the `isMultilingual`.
https://github.com/bis-med-it/pysdmx/blob/5aab14598ff4b5cede2282e1ed9f0a77146fe7d7/src/pysdmx/model/concept.py#L72

Furthermore, in the [Data Type](https://github.com/bis-med-it/pysdmx/blob/5aab14598ff4b5cede2282e1ed9f0a77146fe7d7/src/pysdmx/model/concept.py#L23) the following are missing:
```
Count
InclusiveValueRange
ExclusiveValueRange
Incremental
StandardTimePeriod
BasicTimePeriod
GregorianTimePeriod
ReportingTimePeriod
ReportingYear
ReportingSemester
ReportingTrimester
ReportingQuarter
ReportingMonth
ReportingWeek
ReportingDay
TimeRange
Duration
GeospatialInformation
XHTML
```

sosna:
There are still types that we need to add, as a colleague had an error recently because of some missing types such as ReportingTimePeriod

PR #328, opened 2025-06-03
Support additional facets and data types

sosna:
This PR adds support for the `is_multilingual` facet and also for additional data types. All types that can be set in the FMR are now supported.


sosna:
> @sosna in #88 several data types and facets are specified to be added, but I just see support for multilingual. Do we need to implement as well the remaining ones?

There is one new facet (`is_multilingual`) and 3 new data types (`DURATION`, `GREGORIAN_TIME_PERIOD` and `XHTML`). 

`XHTML` and `is_multilingual` were added so that we have full parity with the features offered in the FMR.

`DURATION` and `GREGORIAN_TIME_PERIOD` were added because they are part of ISO 8601 and we already support all the other ISO 8601 types referenced in SDMX.

I hope this clarifies 😊

javihern98 (approved):
Looks good, thanks 😊

PR #332, opened 2025-06-04
Add SDMX-ML Structures 3.0.0 writer

jfrancma:
Implement the writer for SDMX-ML 3.0 for the following objects

- OrganisationScheme
- AgencyScheme
- Codelists
- ConceptScheme
- DataStructureDefinition
- Dataflows
- TransformationScheme
- RulesetScheme
- UserDefinedOperatorScheme
- VtlMappingScheme
- CustomTypeScheme
- NamePersonalisationScheme

### DIFF
--- a/tests/io/xml/sdmx30/writer/test_structures_writing.py
+++ b/tests/io/xml/sdmx30/writer/test_structures_writing.py
@@ -912,7 +912,8 @@ def test_write_custom_type_scheme(
 def test_datastructure_read_write(
     complete_header, full_datastructure_sample, write_datastructure_sample
 ):
-    # TODO: This test will be revised in #166, currently fails because of Group implementation
+    # TODO: This test will be revised in #166, currently fails
+    #  because of Group implementation
     with pytest.raises(NotImplementedError, match="Group"):
         read_sdmx(full_datastructure_sample, validate=True)
     # Read the SDMX-ML file


javihern98:
Need to merge first https://github.com/bis-med-it/pysdmx/pull/322 before merging this one

sosna:
I have just approved the reader's branch. Could it please be merged with this branch before I start my review? That would help with the testing. Thanks in advance 😊

javihern98:
> I have just approved the reader's branch. Could it please be merged with this branch before I start my review? That would help with the testing. Thanks in advance 😊

Yes, need to fix a couple of conflicts first

javihern98:
@sosna Ready now for your review 😊

sosna (changes requested):
I have started the review by attempting the following:

1. Exporting the NAMAIN_IDC_N dataflow maintained by SDMX including all its descendants, in SDMX-ML 2.1, from the [SDMX Global Registry](https://registry.sdmx.org/).
2. I read this and used the SDMX-ML 3.0 structure writer to rewrite this as SDMX-ML 3.0.
3. I then wanted to use the SDMX-ML 3.0 Structure reader and assert that what I read in point 1 is equal to what I read in point 3. 

This is the code:

```python
from pysdmx.io.xml.sdmx30.writer.structure import write


msg1 = read_sdmx("Dataflow.xml")

write(msg1.structures, "structures.xml")

msg2 = read_sdmx("structures.xml")

assert msg1.structures == msg2.structures
```

However, reading the SDMX-ML 3.0 output fails with the following validation errors:

```console
pysdmx.errors.Invalid: Validation Error: Element '{http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}AgencyScheme', attribute 'version': The attribute 'version' is not allowed.;
Element '{http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}TimeDimension', attribute 'position': The attribute 'position' is not allowed.;
Element '{http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}None': This element is not expected. Expected is one of ( {http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}Dataflow, {http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}Dimension, {http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}Group, {http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}Observation ).
```

So, apparently, the SDMX-ML 3.0 writer outputs invalid structure files. Could you please investigate?

Thanks!

javihern98:
> So, apparently, the SDMX-ML 3.0 writer outputs invalid structure files. Could you please investigate?

@sosna Could you double check if it works? It seems the urns are not working:

```bash
E               Element '{http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}Enumeration': [facet 'pattern'] The value '
E                                                   urn:sdmx:org.sdmx.infomodel.codelist.Codelist=IMF:CL_ORGANISATION(1.5)
E                                               ' is not accepted by the pattern '.+\.codelist\.ValueList=.+'.;
E               Element '{http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure}ConceptIdentity': [facet 'pattern'] The value '
E                                               urn:sdmx:org.sdmx.infomodel.conceptscheme.Concept=ESTAT:CS_NA(1.6).OBS_VALUE
E                                           ' is not accepted by the pattern '.+\.conceptscheme\.Concept=.+'.
```

javihern98:
> I read the [technical note of SDMX-ML 3.0](https://sdmx.org/wp-content/uploads/SDMX_3-1-0_SECTION_6_FINAL.pdf) and it is not clear what we should do in this case.

@sosna I agree with option 4 as it should be really easy to implement, but my suggestion is to simply add a `post_init` check to MaintainableArtefact to ensure isFinal is set to True if version has no `-` in it. If the IM is already SDMX 3.0 and isFinal is related to the version, then this could be also an easy option

sosna:
> @sosna I agree with option 4 as it should be really easy to implement, but my suggestion is to simply add a `post_init` check to MaintainableArtefact to ensure isFinal is set to True if version has no `-` in it. If the IM is already SDMX 3.0 and isFinal is related to the version, then this could be also an easy option

Indeed, we could do both but doing this in the model would affect other parts that might not be expecting that. So, we can revise this later but, at this stage, as I feel it's safer to limit the scope to the deserialization of SDMX-ML 3.0 payloads. Would you be OK with that?

javihern98:
@sosna I implemented the version and is_final handling on the reader, please check it

sosna (changes requested):
Thanks, @javihern98. I believe it is still failing. I have added a test reproducing the issue (which explains all the failing tests in the CI pipeline).

sosna (changes requested):
If you follow the steps indicated in my [previous message](https://github.com/bis-med-it/pysdmx/pull/332#pullrequestreview-2912587912), you will see that the checks still fail for some structure types. 

For example, it fails when comparing the 2 DSDs. Apparently, `COMMENT_DSET` is not equal. All the other components (as well as the DSD properties, like version, agency, etc.) are considered equal. Initial investigations seem to indicate that attachment level is the culprit.

sosna:
> Regarding the attachment level, the only issue we found was to change the attachment level from None in SDMX-ML 2.1 to Observation in SDMX-ML 3.0, as None is not valid in the XSD for SDMX-ML 3.0.

I think I know what's wrong but I need another couple of hours to come up with a test reproducing the issue😊

sosna:
> > > Regarding the attachment level, the only issue we found was to change the attachment level from None in SDMX-ML 2.1 to Observation in SDMX-ML 3.0, as None is not valid in the XSD for SDMX-ML 3.0.
> > 
> > 
> > I think I know what's wrong but I need another couple of hours to come up with a test reproducing the issue😊
> 
> No problem, let me know if you need a quick call to look into it 😊

Thanks! I think I know what's happening. There are at least three issues (maybe more), related to the handling of attributes' attachment level.

Here is the relevant extract of the SDMX-ML 2.1 input (i.e. the starting point):

```xml
<str:Attribute urn="urn:sdmx:org.sdmx.infomodel.datastructure.DataAttribute=ESTAT:NA_MAIN(1.6).COMMENT_DSET" assignmentStatus="Conditional" id="COMMENT_DSET">
  <str:ConceptIdentity>
    <Ref maintainableParentID="CS_NA" package="conceptscheme" maintainableParentVersion="1.6" agencyID="ESTAT" id="COMMENT_DSET" class="Concept"/>
  </str:ConceptIdentity>
  <str:LocalRepresentation>
    <str:TextFormat textType="String" maxLength="1050"/>
  </str:LocalRepresentation>
  <str:AttributeRelationship>
    <str:None/>
  </str:AttributeRelationship>
</str:Attribute>
```

As you can see, the attribute relationship is set to None, which, prior to SDMX 3.0, means that the attribute is attached to the dataset/dataflow level. However, when using pysdmx to deserialize the XML file, the generated attribute does not contain any attachment level. This is wrong, i.e. the attachment-level should be set to `D`. This is the first issue.

As a matter of fact, I don't think it should even be possible to instantiate an attribute object without specifying an attachment level, so maybe we should add a `__post_init__` check? This is the second issue.

Then, the third issue is that, when writing the model objects to SDMX-ML 3.0, the inexistent (i.e. None) attribute level gets translated into `<Observation/>`.

```xml
<str:Attribute usage="optional" id="COMMENT_DSET" urn="urn:sdmx:org.sdmx.infomodel.datastructure.DataAttribute=ESTAT:NA_MAIN(1.6).COMMENT_DSET">
    <str:ConceptIdentity>urn:sdmx:org.sdmx.infomodel.conceptscheme.Concept=ESTAT:CS_NA(1.6).COMMENT_DSET</str:ConceptIdentity>
    <str:LocalRepresentation><str:TextFormat maxLength="1050" textType="String"/></str:LocalRepresentation>
    <str:AttributeRelationship><str:Observation/></str:AttributeRelationship>
</str:Attribute>
```

This is wrong, as, in essence, we are attaching a dataset-level attribute to the observation-level. Instead, it should be attached the dataflow level:

```xml
<str:Attribute urn="urn:sdmx:org.sdmx.infomodel.datastructure.DataAttribute=ESTAT:NA_MAIN(1.6).COMMENT_DSET" usage="optional" id="COMMENT_DSET">
  <str:ConceptIdentity>urn:sdmx:org.sdmx.infomodel.conceptscheme.Concept=ESTAT:CS_NA(1.6).COMMENT_DSET</str:ConceptIdentity>
  <str:LocalRepresentation>
    <str:TextFormat textType="String" maxLength="1050"/>
  </str:LocalRepresentation>
  <str:AttributeRelationship>
    <str:Dataflow/>
  </str:AttributeRelationship>
</str:Attribute>
```

This is why the equality check failed. For the 2.1 message, the attachment-level was set to None, which is wrong, while for the 3.0 message, it was set to observation, which is wrong too.

Is it OK for you if I create 3 new issues (bugs) and assign them to the 1.4.0 release?

javihern98:
Hi @sosna, yes, let´s generate 3 separated issues and later on merge this branch. I believe neither of them are related to writing structures but the readers or the model itself.

javihern98:
@sosna I added a test for issue 3 and refactored the code for attribute relationship to remove the check on attachment_level=None, checks should pass once we merge #344

sosna (approved):
Nice work, thanks 😊!

I have run roundtrip tests and also validated the generated outputs and it all went well.

Issue #166, opened 2024-12-23
Add support for Groups in SDMX-ML Data readers and writers


javihern98:
Add support for GroupDimensions in Schema and data readers and writers

javihern98:
@sosna @stratosn We are planning to support as well the Groups in data messages on SDMX-ML 2.1. We are currently struggling to see how we can include the Group class in the model, as it cannot be a Component class but somehow we must add this information to the Schema (and DataFlow, DataStructureDefinition, etc).

sosna:
Thanks, @javihern98, and happy new year :).

Is this something that can be discussed after the 1.0.0 release?

javihern98:
Sure! We can implement it afterwards, just to have it in the backlog for future releases

javihern98:
As agreed on meeting with @sosna , we need to discuss with @stratosn on how we can include the GroupDimension into the DataStructureDefinition.

We need to check the deprecation status of groups in SDMX 3.x

javihern98:
Agreed on the meeting:

- Implement the Group, it is needed for both data reading and writing
- We cannot push down the dimensions and attributes related to a group into the Series level, unless the Dataset does not have the Schema/DataStructureDefinition object at its structure
- The Group definition will have the list of dimension codes of the groups. 
- If an attribute is attached to a group (as in the attachment level), we will add the list of dimension codes in the attachment level. 

When writing, the attributes attached to a list of dimensions that are also included in a group (all of the dimensions codes in the attachment level must be present in the group), then we will write the attributes at the Group level.

javihern98:
> Dear all, we have internal requests now for reading DSDs with groups (also in 3.0). I assume (considering the draft PR) that this is still feasible for the next version, or? (maybe not by the end of July, but soon after)

Dear @stratosn, this issue is ready from our side, we are simply awaiting for the release of 1.4.0 to start the review process. This will be present in 1.5.0, hoping we can release it before the end of August (depends only on holidays and availability from both sides). BIS colleagues who request it can always checkout the related branch `cr-166` and start some testing to speed up the process if it is very urgent.

PR #344, opened 2025-06-24
Remove optionality for attachment level on attributes

javihern98:
- Added `__post_init__` check to ensure attachment level is set when the Component Role is Attribute
- Fixed SDMX-ML reader to ensure the default relationship is Dataset (i.e. when Attribute relationship is None)
- Fixed tests that defined attributes without an attachment level
- Added tests for Group and AttachmentGroup relationship (to be modified in #166 
- Fixed Pandas warning in groupby.apply at Generic Data writer (out of scope, but minimal change)


javihern98:
Need to add a roundtrip test on SDMX-ML 2.1 for this issue

jfrancma (commented):
Looks good, ready for BIS review.

src/pysdmx/io/xml/__structure_aux_reader.py
```diff
@@ -551,8 +551,8 @@ def __format_con_id(self, concept_ref: Dict[str, Any]) -> Dict[str, Any]:
         return rep
 
     @staticmethod
-    def __format_relationship(json_rel: Dict[str, Any]) -> Optional[str]:
-        att_level = None
+    def __format_relationship(json_rel: Dict[str, Any]) -> str:
+        att_level = "D"
 
         for scheme in [DIM, PRIM_MEASURE, MEASURE, DFW, OBSERVATION]:
```
sosna, line 557: These tokens (DIM, DFW, etc.) seem to be imported from pysdmx.io.xml.sdmx21? Is that the case? If so, this is odd in a file that should be version-agnostic, isn't it? I mean, maybe the structure file being read is an SDMX-ML 3.0?

sosna, line 557: What is meant with `scheme` here? It reminds of SDMX item scheme, but I guess it's something else, as this would not be relevant to derive the attachment level?

src/pysdmx/io/xml/__structure_aux_reader.py
```diff
@@ -551,8 +551,8 @@ def __format_con_id(self, concept_ref: Dict[str, Any]) -> Dict[str, Any]:
         return rep
 
     @staticmethod
-    def __format_relationship(json_rel: Dict[str, Any]) -> Optional[str]:
-        att_level = None
+    def __format_relationship(json_rel: Dict[str, Any]) -> str:
```
sosna, line 554: If this method is meant to return an attachment level, maybe the name __format_relationship is confusing? What about something like __get_attachment_level instead?

sosna, line 554: Also, I'm not sure what is the intent to call an attribute json_rel in a module dealing with XML parsing. What does this attribute hold?

src/pysdmx/io/xml/__structure_aux_reader.py
```diff
@@ -551,8 +551,8 @@ def __format_con_id(self, concept_ref: Dict[str, Any]) -> Dict[str, Any]:
...
+        att_level = "D"
```
sosna, line 555: I don't think we should have a default attachment level (see suggestion below, in the if/else clauses)

src/pysdmx/io/xml/__structure_aux_reader.py
sosna: In the if/elif/else clause, I think we should change 2 things:

- We should check if the attachment level is set to group and, if so, we should throw an Unimplemented error (we don't yet support groups)
- The else clause should throw an Invalid error (any attachment level which is not one of the expected ones is probably invalid?)

sosna: The names of the attachment levels have changed with the SDMX versions. Has this been taken into account?

sosna (changes requested):
I have made a few comments/questions. Some of these might be due to a misunderstanding on my side, so happy to talk if necessary 😊. Also, do we have enough tests on the readers side (meaning, do we have all possible options in SDMX-ML 2.1 and SDMX-ML 3.0 input files used in the tests)?

src/pysdmx/io/xml/__structure_aux_reader.py
javihern98: See below comment in if/then clauses

sosna:
> I implemented it making the attachment_level="D" at the else clause due to the impossibiliity of testing it with an Invalid error using any SDMX-ML 2.1/3.0 Structures file. The only way would be to use a file that is not compliant with the XSD Schemas, which is against the basis of the Structures parsers design (we are only covering whatever is in the XSD Schemas for SDMX-ML)

Thanks, @javihern98, you are right! I had not considered the schema-based validation.

PR #346, opened 2025-06-30
GDS connector API reference docs

javihern98:
# Summary of changes

- Refactored docstrings on GDSClient
- Added API reference for GDS client and GDS model classes


docs/api/gds/async.rst
```diff
@@ -0,0 +1,24 @@
...
+>>> import asyncio
+>>> from pysdmx.api.gds import AsyncGdsClient
+>>> async def main():
+>>>     gr = AsyncGdsClient()
+>>>     gds_agencies = await gr.get_agencies("BIS")
+>>>     print(gds_agencies)
+>>> asyncio.run(main())
+
+.. important::
+    This client is designed to be used with Python's `asyncio` library.
```
sosna, line 16: Maybe we could add that this is the recommended client to be used when using asynchronous frameworks like `fastapi`?

src/pysdmx/api/gds/__init__.py
```diff
@@ -151,36 +151,30 @@ def get_catalogs(
...
-                  (or not) the artefacts referenced by the
-                  artefact to be returned.
-
-                    Option none: No referenced artefacts will be returned.
-                    Option children: Returns the artefacts
-                      referenced by the artefact to be returned.
-
+            catalog: The agency maintaining the catalog.
+            resource: The resource ID(s) to query. Defaults to '*'.
+            version: The version(s) of the resource. Defaults to '*'.
```
sosna, line 156: Is that so? In SDMX-REST, by default, you don't get all versions but only the latest one...

src/pysdmx/api/gds/__init__.py
```diff
@@ -202,8 +196,7 @@ def get_sdmx_apis(
         """Get the list of SDMX API versions.
```
sosna, line 196: Does it really return the APIS versions? I have the impression it returns the APIs, optionally filtered by version.

src/pysdmx/api/gds/__init__.py
```diff
@@ -202,8 +196,7 @@ def get_sdmx_apis(
...
 
         Args:
-            api_version: The ID of the SDMX API version
-            to query. Defaults to '*'.
+            api_version: The SDMX_API to query. Defaults to '*'.
```
sosna, line 199: Should we say something like: "The version of the SDMX API to be returned." instead of "The SDMX_API to query."?

sosna (dismissed):
It reads very nicely, thanks 😊.

I have a few minor comments / questions, but as they are all minor, I have approved the PR already.

In general, I'm wondering whether we should not give a hint about when to use a sync vs. an async client. For example, sync could be favored for simplicity or when the application is procedural (i.e. we need to wait before proceeding to the next step), while async could be favored when using async frameworks (like the fastapi), when writing non-blocking applications, or when there is a wish to execute multiple requests in parallel.

src/pysdmx/api/gds/__init__.py
javihern98: In this case yes, as the SDMX API ID is in fact the SDMX-REST version, any suggestion on the wording to use?

javihern98:
@sosna I added the note on FastAPI with an example code and changed wording. From my side no more changes are needed, let me know if I can resolve the conversation and merge 😊

javihern98:
It seems it requires again your approval after the small changes @sosna 😅

PR #349, opened 2025-07-02
Fixed null handling on data writers

javihern98:
- Ensure we convert types and replace null values to empty strings
- Added specific test based on issue comments

### DIFF
--- a/src/pysdmx/model/message.py
+++ b/src/pysdmx/model/message.py
@@ -84,6 +84,14 @@ class Header(Struct, repr_omit_defaults=True, kw_only=True):
     structure: Optional[Dict[str, str]] = None
     dataset_id: Optional[str] = None
 
+    def __post_init__(self) -> None:
+        """Header post-initialization."""
+        if isinstance(self.sender, str):
+            self.sender = Organisation(id=self.sender)
+
+        if isinstance(self.receiver, str):
+            self.receiver = Organisation(id=self.receiver)
+
     def __str__(self) -> str:
         """Custom string representation without the class name."""
         processed_output = []
--- a/tests/model/test_message.py
+++ b/tests/model/test_message.py
@@ -244,6 +244,24 @@ def test_message_str():
     assert s == expected_str
 
 
+def test_header_initialization_with_strings():
+    header = Header(
+        id="12345",
+        test=True,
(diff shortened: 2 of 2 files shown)


sosna (commented):
How would this work in the following scenario?

- The input is SDMX-ML. In there, we have `OBS_VALUE` explicitely set to `NaN`. 
- The output is SDMX-CSV. 

According to the [SDMX-CSV spec](https://github.com/sdmx-twg/sdmx-csv/blob/master/data-message/docs/sdmx-csv-field-guide.md#intentionally-missing-values), we should find either `NaN` or `#NA` depending on the type.

Will this be the case and, if so, do we have a test to prove it?

javihern98:
I believe this will not be the case, I missed that part of the specification. We need to add a test as well on these values

jfrancma:
Added tests to check correct spelling of NaN and #NA for CSV

sosna:
I was trying to add a roundtrip test, to make sure explicit NaN gets preserved.

Here it was the start:

```python
from pathlib import Path

import pytest

@pytest.fixture
def input():
    base_path = Path(__file__).parent / "samples" / "ecb_exr.xml"
    return str(base_path)

def test_nan_roundtrip_ml21_csv2(input):
    # Read the SDMX XML data
    data = read_sdmx(input, validate=True).data

    # Write to SDMX-CSV 2.0
    result = write_sdmx(data, Format.DATA_SDMX_CSV_2_0_0)
```

An error is thrown on the last line. 

If I add a path, it works:

```python
from pathlib import Path

import pytest

@pytest.fixture
def input():
    base_path = Path(__file__).parent / "samples" / "ecb_exr.xml"
    return str(base_path)

def test_nan_roundtrip_ml21_csv2(input):
    # Read the SDMX XML data
    data = read_sdmx(input, validate=True).data

    # Write to SDMX-CSV 2.0
    result = write_sdmx(data, Format.DATA_SDMX_CSV_2_0_0, "output.csv")
    print(result)
```

According to the documentation, that shouldn't be the case, isn't it?

Could you please investigate? Thanks

javihern98:
Hi @sosna that error is already solved in #355

sosna (approved):

PR #357, opened 2025-07-03
Roundtrip reading SDMX-ML 2.1 and writing SDMX-ML 3.0 raises errors due to Agency URNs

jfrancma:
- Fix URN write for Agency 3.0
- Remove isFinal write for AgencyScheme 2.1


javihern98 (commented):
Looks good, ready for BIS review😊

tests/io/xml/sdmx30/writer/samples/bis_der_agencies.xml
sosna, line 1: There might be a pending issue (discussed bilaterally) re. the content of this file.

src/pysdmx/io/xml/__structure_aux_writer.py
```diff
@@ -300,7 +303,7 @@ def __write_maintainable(
         f" isExternalReference="
         f"{str(maintainable.is_external_reference).lower()!r}"
     )
-    if not references_30:
+    if not references_30 and not (isinstance(maintainable, AgencyScheme)):
```
sosna, line 306: We should test with the other organisation schemes too. I suspect the same rule applies, for example, to DataProviderScheme.

javihern98: There is no test with DataProviderScheme on SDMX-ML reading and writing as it is not implement, therefore we should add the support for it in a separate issue, do you agree?

PR #359, opened 2025-07-03
Raise an error when trying to write a DataStructureDefinition with more than one measure in SDMX-ML 2.1

javihern98:
# Summary of changes

- Added an error message to prevent writing SDMX-ML 2.1 when a DataStructureDefinition has multiple measures
- Refactored name of variable in write_components

### DIFF
--- a/src/pysdmx/io/xml/__structure_aux_writer.py
+++ b/src/pysdmx/io/xml/__structure_aux_writer.py
@@ -381,7 +381,7 @@ def __write_item(
 
 
 def __write_components(
-    item: DataStructureDefinition, indent: str, references_30: bool = False
+    dsd: DataStructureDefinition, indent: str, references_30: bool = False
 ) -> str:
     """Writes the components to the XML file."""
     outfile = f"{indent}<{ABBR_STR}:{DSD_COMPS}>"
@@ -392,13 +392,19 @@ def __write_components(
         PRIM_MEASURE: [],
     }
 
-    for comp in item.components:
+    for comp in dsd.components:
         if comp.role == Role.DIMENSION:
             components[DIM].append(comp)
         elif comp.role == Role.ATTRIBUTE:
             components[ATT].append(comp)
         else:
             components[PRIM_MEASURE].append(comp)
+
+    if not references_30 and len(components[PRIM_MEASURE]) > 1:
+        raise Invalid(
+            f"SDMX-ML 2.1 does not support multiple measures. "
+            f"Check the {dsd.short_urn}."
+        )
 
     position = 1
     for _, comps in components.items():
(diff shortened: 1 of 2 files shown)


src/pysdmx/io/xml/__structure_aux_writer.py
```diff
@@ -392,14 +392,20 @@ def __write_components(
...
+    for comp in dsd.components:
         if comp.role == Role.DIMENSION:
             components[DIM].append(comp)
         elif comp.role == Role.ATTRIBUTE:
             components[ATT].append(comp)
         else:
             components[PRIM_MEASURE].append(comp)
 
+    if not references_30 and len(components[PRIM_MEASURE]) > 1:
+        raise Invalid(
```
sosna, line 404: pysdmx errors may have up to 3 fields (title, description, csi). So, in this case, we could have a generic title (like "Request cannot be fulfilled"), followed by a description (like the string you compute). And then, as csi, I would maybe add a property with the measures found? For example:

```python
{
    "measures_found": ["TURNOVER", "OPEN_INTEREST", "YEAR_CHANGE"]
}
```

What do you think?

javihern98: I agree, this is cleaner. Let me add it to the error message

src/pysdmx/io/xml/__structure_aux_writer.py
```diff
@@ -394,14 +394,24 @@ def __write_components(
...
     }
 
-    for comp in item.components:
+    for comp in dsd.components:
         if comp.role == Role.DIMENSION:
             components[DIM].append(comp)
         elif comp.role == Role.ATTRIBUTE:
             components[ATT].append(comp)
         else:
             components[PRIM_MEASURE].append(comp)
```
sosna, line 403: I have just noticed: `PRIM_MEASURE` is a bit confusing as name. The reason is that the primary measure, in SDMX, meant `OBS_VALUE`. But now, in this scenario (multiple measures), it won't be the case. Therefore, maybe we can align to the way things are done for dimensions (`DIM`) and attributes (`ATT`), and we simply call it `MSR`?

sosna (approved):
Thanks, @javihern98. I have made a minor comment, but I already approved the PR, just in case you disagree with it.

src/pysdmx/io/xml/__structure_aux_writer.py
javihern98: Yes, but it has some further implications in Role_MAPPING and other parts. We do need to adapt the code to SDMX 3.1 in next release, so I find it best to rename it there. I will add a comment in the issue

Issue #327, opened 2025-06-03
Add support for SDMX-ML 3.1


javihern98:
- Add namespace for SDMX 3.1
- Review data and structure readers and writers and add tests

javihern98:
Following discussion in https://github.com/bis-med-it/pysdmx/pull/359#discussion_r2184701238, we should rename in the Structures writers any reference to PRIM_MEASURE to use simply MEASURE (or a new token MSR, for consistency with DIM and ATT), as the model is based in SDMX 3.0

PR #366, opened 2025-07-21
Add support for SDMX-ML 3.1

jfrancma:
- Add 3.1 Structure and Structure specific readers.
- Add 3.1 Structyre and Structure specific writers.
- Add 3.1 readers and writers to general reader and writer.


javihern98 (commented):
Reduce size of tests and use same fixtures for writing on all SDMX-ML versions. Need to implement https://github.com/bis-med-it/pysdmx/issues/327#issuecomment-3034951998

javihern98 (commented):
Looks good 😊 Ready for BIS review on 1.5.0

src/pysdmx/io/format.py
```diff
@@ -19,20 +19,25 @@ class Format(Enum):
         f"{_BASE}structurespecifictimeseriesdata+xml;version=2.1"
     )
     DATA_SDMX_ML_3_0 = f"{_BASE}data+xml;version=3.0.0"
+    DATA_SDMX_ML_3_1 = f"{_BASE}data+xml;version=3.1"
```
sosna, line 22: I believe it should be 3.1.0 but let me check with the TWG

sosna (commented):
I need to check the expected version of the media type for the format, i.e. 3.1 vs. 3.1.0.

Also, would it be possible to add the most important sample files published with the release: https://github.com/sdmx-twg/sdmx-ml/tree/master/samples? 

I think this was already discussed for the 1.4.0 release and we agreed it would be a nice thing to do, going forward? If so, we could focus on the most important files, such as the core artefacts (codelists, concept schemes, DSDs, etc.) and the EXR data files.

tests/api/qb/registration/test_registry_format.py
```diff
@@ -5,6 +5,7 @@ def test_expected_formats():
     expected = [
         "application/vnd.sdmx.registry+xml;version=2.1",
         "application/vnd.sdmx.registry+xml;version=3.0",
```
sosna, line 7: According to the spec, it should be 3.0.0

sosna (changes requested):

tests/api/qb/registration/test_registry_format.py
jfrancma: changed to 3.0.0

sosna (approved):

PR #381, opened 2025-09-01
Add support for SDMX-CSV 2.1 data message

jfrancma:
- Refactor csv readers and writers
- Implemented Optional parameters (Labels, Time format and Keys)
- Added csv 2.1 to general reader and writer
- The toolkit tests were divided into two: vtl and pd.
- Refactor get_codes to toolkit/pd to avoid import of xml in csv writer

### DIFF
--- a/src/pysdmx/io/csv/__csv_aux_writer.py
+++ b/src/pysdmx/io/csv/__csv_aux_writer.py
@@ -4,10 +4,9 @@ from typing import List, Literal, Optional, Sequence
 import pandas as pd
 
 from pysdmx.io.pd import PandasDataset
-from pysdmx.io.xml.__write_data_aux import get_codes
 from pysdmx.model import Schema
 from pysdmx.model.dataset import ActionType
-from pysdmx.toolkit.pd._data_utils import format_labels
+from pysdmx.toolkit.pd._data_utils import format_labels, get_codes
 
 SDMX_CSV_ACTION_MAPPER = {
     ActionType.Append: "A",
--- a/src/pysdmx/io/xml/__write_data_aux.py
+++ b/src/pysdmx/io/xml/__write_data_aux.py
@@ -1,6 +1,4 @@
-from typing import Any, Dict, List, Optional, Sequence, Tuple
-
-import pandas as pd
+from typing import Dict, Optional, Sequence
 
 from pysdmx.errors import Invalid
 from pysdmx.io.pd import PandasDataset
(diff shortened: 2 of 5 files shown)


javihern98 (commented):

src/pysdmx/model/dataset.py
```diff
@@ -25,6 +25,7 @@ class ActionType(Enum):
     Replace = "Replace"
     Delete = "Delete"
     Information = "Information"
+    Merge = "Merge"
```
sosna, line 28: This sounds wrong to me. According to the SDMX-CSV doc, Merge is the same as Append, meaning Merge is the new name and Append has been deprecated. Therefore, I don't think we need a new action in the model. Instead, what we need, is to be able to write the "correct" action in SDMX messages, depending on their version (Append for old versions and Merge for new ones).

src/pysdmx/io/csv/__csv_aux_reader.py
```diff
@@ -0,0 +1,89 @@
...
+from pysdmx.errors import Invalid
+from pysdmx.io.pd import PandasDataset
+from pysdmx.model.dataset import ActionType
+
+ACTION_SDMX_CSV_MAPPER_READING = {
+    "A": ActionType.Append,
+    "D": ActionType.Delete,
+    "R": ActionType.Replace,
+    "I": ActionType.Information,
+    "M": ActionType.Merge,
```
sosna, line 12: Please, cf. comment made in pysdmx.model.dataset.py

src/pysdmx/io/csv/__csv_aux_writer.py
```diff
@@ -0,0 +1,120 @@
...
+from pysdmx.model import Schema
+from pysdmx.model.dataset import ActionType
+from pysdmx.toolkit.pd._data_utils import format_labels
+
+SDMX_CSV_ACTION_MAPPER = {
+    ActionType.Append: "A",
+    ActionType.Replace: "R",
+    ActionType.Information: "I",
+    ActionType.Delete: "D",
+    ActionType.Merge: "M",
```
sosna, line 17: Please, cf. comment made on pysdmx.model.dataset.py

src/pysdmx/model/dataset.py
sosna: I disagree on this. The domain model is supposed to be neutral and focus on what is relevant for the business. On the business side, you have a merge vs. a replace. The integration logic will be different depending on the selected action. This is what matters. Now, if in the past, merge used to be called append, it doesn't matter, it's still 2 actions on the integration side.

javihern98: Model is based on SDMX 3.x, I suggest to keeep it as long as we have a check that changes Append to Merge when we are writing SDMX-CSV 2.1

sosna: What do you mean with "I suggest to keep it"? Keeping the new merge action, or keeping the old enum (with only 4 values)?

By the way, just for reference, the SDMX 3.1 Information Model states the following:

"The ActionType enumeration is used to specify the action that a receiving system should take when processing the content that is the object of the action. It is enumerated as follows:

- Append: Data or metadata is an incremental update for an existing data/metadata set or the provision of new data or documentation (attribute values) formerly absent. If any of the supplied data or metadata is already present, it will not replace that data or metadata. This corresponds to the "Update" value found in version 1.0 of the SDMX Technical Standards.
- Replace: Data/metadata is to be replaced and may also include additional data/metadata to be appended.
- Delete: Data/Metadata is to be deleted.
- Information: Data and metadata are for information purposes."

jfrancma: I'm a little confused that the information model doesn't mention merge and that the CSV 2.1 field guide replaces information and append with merge. 
But with what @sosna mentioned, I think it makes sense to leave the actions we have and replace append with merge when writing and merge with append when reading.

sosna: Perfect! Thanks, @jfrancma 😊

sosna (approved):
Thanks, this looks nice 😊

PR #385, opened 2025-09-02
Add a workflow for extra dependencies

jfrancma:


src/pysdmx/toolkit/vtl/_validations.py
```diff
@@ -62,7 +61,7 @@ def _ruleset_scheme_validations(ruleset_scheme: RulesetScheme) -> None:
 
 def _ruleset_scheme_reference_validations(ruleset_scheme: Reference) -> None:
     """Additional validation checks for ruleset schemes."""
-    if ruleset_scheme.sdmx_type != RULE_SCHEME:
+    if ruleset_scheme.sdmx_type != "RulesetScheme":
```
javihern98, line 64: Why is this change needed here? Is this to prevent errors with the extras handling? If so, please add a note to the summary of changes

src/pysdmx/toolkit/vtl/_validations.py
```diff
@@ -105,7 +104,7 @@ def _user_defined_operator_scheme_validations(
 
 def _udo_scheme_reference_validations(udo_scheme: Reference) -> None:
     """Additional validation checks for ruleset schemes."""
-    if udo_scheme.sdmx_type != UDO_SCHEME:
+    if udo_scheme.sdmx_type != "UserDefinedOperatorScheme":
```
javihern98, line 107: Same as above

tests/conftest.py
```diff
@@ -0,0 +1,73 @@
+from pathlib import Path
+
+import pytest
+
+PATH_RULES = {
+    "/tests/io/xml/": ("xml", True),
```
javihern98, line 6: Some tests here require both xml and data extras, like reading SDMX-ML Structure Specific. How is this handled?

.github/workflows/extra.yml
```diff
@@ -0,0 +1,84 @@
...
+            extras_to_install: ""
+          - extra: vtl
+            marker: vtl
+            extras_to_install: "vtl"
+          - extra: data
+            marker: data
+            extras_to_install: "data"
+          - extra: xml
+            marker: xml
+            extras_to_install: "xml data"
```
javihern98, line 34: Could we have another one with only xml extra and no data extra?

javihern98 (commented):
Some questions, base looks really good, just to confirm that we are testing everything correctly

src/pysdmx/toolkit/vtl/_validations.py
jfrancma: This change was made to avoid the xml amount in vtl.

tests/conftest.py
jfrancma: The tests for testing XML and XML with data were separated.

sosna (commented):
This is brilliant, thanks a lot 😊! Hopefully, this solves our dependencies issue once and for all!

I just made a few minor comments, but, overall, this is very good!

tests/io/xml/sdmx30/reader/test_reader.py
```diff
@@ -90,6 +89,10 @@ def test_prov_agree_30_groups_series(samples_folder):
 
 
 def test_data_no_structure_specific(samples_folder):
```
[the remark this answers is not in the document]
jfrancma: Since this test imports structure_specific, it also needs the extra data, which is why the xml marker is not added.

pytest.ini
```diff
@@ -0,0 +1,8 @@
+[pytest]
+testpaths = tests
+markers =
+    xml: SDMX-ML structures readers and writers.
+    xmlWithData: SDMX-ML readers and writers with data.
+    data: SDMX-CSV and read_sdmx tests
+    vtl: VTL tests
+    noextra: Tests for dependencies not requiring extra packages
```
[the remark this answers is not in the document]
jfrancma: The API tests have been separated into two groups: those that do not require extra and those that require extra dc.

PR #412, opened 2025-10-02
Extract concept urn directly parsing the JSON concept

### LINKED ISSUE #410: Extract concept urn when reading SDMX-JSON structure files
The urn concept is not extracted at the moment when reading SDMX-JSON structure files. This means that, at the moment, it is not possible to read a file in SDMX-JSON and then write it back, as the urn is mandatory information when writing.

sosna:
The concept urns where added to the concept when parsing the concept scheme. This worked in most cases, but also meant that urn would not be available when parsing the concept directly (like, for example, when associating a component to its underlying concept). This PR addresses the issue by making sure the urn will always be available present in the concept extracted from JSON.

Close #410


sosna:
Indeed, we have a test that urns have been extracted when reading concepts, either as SDMX-JSON or as Fusion-JSON: https://github.com/bis-med-it/pysdmx/blob/cr-410/tests/api/fmr/concept_checks.py#L55

Besides, for the read/write roundtrip, my understanding is that this test verifies that what we wrote is equivalent to what we read: https://github.com/bis-med-it/pysdmx/blob/9d82d8277224964b36aac724dd5d517497d94036/tests/io/test_general_writer.py#L145

In my understanding, this is sufficient. Would you agree?

jfrancma (approved):
Looks good!

PR #413, opened 2025-10-02
Add http2 extra to httpx dependency

### LINKED ISSUE #408: Add http2 extra to the httpx dependency
http/2 provides many benefits overt http/1 and its usage should be transparent, i.e. the httpx client is smart enough to know which one to use depending on the capabilities expressed by the server.

sosna:
The http2 extra has been added to the httpx dependency. 

For the rationale, please see the related issue.

Close #408


jfrancma (approved):
Looks good!

PR #428, opened 2025-10-09
Add validation function for JSON files

jfrancma:
- Add validate_sdmx_json for JSON files validation.
- Added a boolean flag to readers to trigger JSON validation (default behaviour in read_sdmx and get_datasets)
- Added `json` extra with `sdmxschemas` and `jsonschema` libraries for SDMX-JSON structural validation
- Edited documentation to reflect these changes in read_sdmx, get_datasets and IO structure tutorial
- (Out of scope) Added extra check on xml parsing as it was missing

### DIFF
--- a/tests/io/test_general_writer.py
+++ b/tests/io/test_general_writer.py
@@ -11,12 +11,12 @@ from pysdmx.io.pd import PandasDataset
 from pysdmx.io.writer import write_sdmx
 from pysdmx.model import Component, Components, Concept, Role, Schema
 
-CSV_1_0_PATH = Path(__file__).parent / "csv" / "sdmx10" / "reader"
-CSV_2_0_PATH = Path(__file__).parent / "csv" / "sdmx20" / "reader"
-XML_2_1_PATH = Path(__file__).parent / "xml" / "sdmx21" / "reader"
-XML_3_0_PATH = Path(__file__).parent / "xml" / "sdmx30" / "reader"
-XML_STR_PATH = Path(__file__).parent
-JSN_2_0_PATH = Path(__file__).parent.parent / "api" / "fmr"
+CSV_1_0_PATH = Path(__file__).parent / "csv" / "sdmx10" / "reader" / "samples"
+CSV_2_0_PATH = Path(__file__).parent / "csv" / "sdmx20" / "reader" / "samples"
+XML_2_1_PATH = Path(__file__).parent / "xml" / "sdmx21" / "reader" / "samples"
+XML_3_0_PATH = Path(__file__).parent / "xml" / "sdmx30" / "reader" / "samples"
+XML_STR_PATH = Path(__file__).parent / "samples"
+JSN_2_0_PATH = Path(__file__).parent.parent / "api" / "fmr" / "samples"
 
 DIMENSIONS = [
     "FREQ",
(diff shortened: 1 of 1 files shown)


tests/io/test_general_reader.py
```diff
@@ -394,7 +402,7 @@ def test_get_json2_structure(sdmx_json_structure):
 
 
 def test_get_json2_refmeta(sdmx_json_refmeta):
-    msg = read_sdmx(sdmx_json_refmeta)
+    msg = read_sdmx(sdmx_json_refmeta, validate=False)
```
javihern98: @sosna We need to review this file as it is not compliant with the SDMX-JSON 2.0 Schema, based on the current implementation for this validation

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "httpx[http2]>=0.*",
     "msgspec>=0.*",
     "parsy>=2.1",
+    "sdmxschemas>=1.0.0"
```
sosna: We agreed to keep the core install as small as possible. So why are the sdmxschemas no longer optional?

sosna (changes requested):
It's good overall, but I would hope we can keep sdmxschemas optional

pyproject.toml
javihern98: Yes, I believe sdmx-schemas should be part of the core dependencies as long as they are required in the XML files but also in the JSON files by default. 

@sosna This library is maintained by us and at some point it should be even part of the SDMX standard, with no further subdependencies and a limited code base. 

You can find the project here: https://github.com/Meaningful-Data/sdmx-schemas

tests/io/test_general_reader.py
sosna: Perfect, thanks, @javihern98. And, good to know about the reviewer / last pusher rule 😊

pyproject.toml
sosna: @javihern98: But how does it differ from the current approach with xml? In our [documentation](https://py.sdmx.io/howto/io_data.html#sdmx-data-io-operations), we clearly indicate that extras are required, and these extras then take care of installing sdmxschemas, isn't it? Why don't we do the same for json?

Alternatively, we could have a new extra `valid-schema` that would take care of installing the schemas in case the user wants to perform schema-based validation. But the issue I see here is that validation is enabled by default, and I'm now wondering whether this is a reasonable default or whether validation should be an opt-in feature... But maybe we shouldn't start this discussion 😂. 

In any case, we are not big fans of schema-based validation and, as a matter of fact, we don't use that internally (we have our own validation function that performs the equivalent of schema-based validation plus much more)...

javihern98: Implemented as defined above, with changes as well on the CI/CD extras workflow to ensure the validation works if only the json extra is installed.

src/pysdmx/io/xml/__parse_xml.py
```diff
@@ -77,6 +78,7 @@ def parse_xml(
     Raises:
         Invalid: If the SDMX data cannot be parsed.
     """
+    __check_xml_extra()
```
javihern98: I found this was missing, but it is needed to ensure the user does not attempt to read an SDMX-ML file without the proper extra and prevent unhandled exceptions 😊 . I added it to the PR summary

sosna: Nice catch 😊

.github/workflows/extra.yml
```diff
@@ -67,9 +70,10 @@ jobs:
           selected = "${{ matrix.extras_to_install }}".split()
           WANT = {
             "data": ["pandas","dateutil"],
+            "json": ["sdmxschemas", "jsonschema"],
             "xml":  ["lxml","sdmxschemas","xmltodict"],
             "xml-data":  ["lxml","sdmxschemas","xmltodict","pandas","dateutil"],
-            "vtl":  ["vtlengine","pandas","dateutil"],
+            "vtl":  ["vtlengine","pandas","dateutil", "jsonschema"],
```
sosna: I'm sure it's correct but, just out of curiosity: For the vtl extra, we need a dependency on jsonschema, but not on sdmxschemas and not on XML?

tests/api/fmr/samples/refmeta/report.json
```diff
@@ -30,70 +30,29 @@
...
                 "agencyID": "BIS.MEDIT",
                 "isExternalReference": false,
-                "isFinal": false,
                 "metadataflow": "urn:sdmx:org.sdmx.infomodel.metadatastructure.Metadataflow=BIS.MEDIT:DTI(1.0)",
                 "targets": [
                     "urn:sdmx:org.sdmx.infomodel.datastructure.Dataflow=BIS.MACRO:BIS_MACRO(1.0)"
                 ],
                 "attributes": [
                     {
-                        "links": [
```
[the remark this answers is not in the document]
javihern98: Need to check

tests/api/fmr/samples/refmeta/report.json
```diff
@@ -19,7 +19,7 @@
                         "rel": "self",
                         "type": "metadataset",
                         "uri": "https://raw.githubusercontent.com/sdmx-twg/sdmx-json/develop/structure-message/tools/schemas/2.0.0/sdmx-json-structure-schema.json",
-                        "urn": "urn:sdmx:org.sdmx.infomodel.metadatastructure.MetadataSet=BIS.MEDIT:DTI_BIS_MACRO(1.0.42)"
```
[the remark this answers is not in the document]
javihern98: validate=False has been added only when having SDMX-JSON REFMETA, at least there is one test that validates the output of an SDMX-JSON. Individual validation tests were added in a previous issue

sosna (approved):

PR #433, opened 2025-10-16
Add support for data structure queries in the FMR clients

### LINKED ISSUE #427: Add support for data structure queries to FMR client
The FMR client does not support retrieving data structures (and their descendants). This went unnoticed so far because of the get_dataflow_info method but it would be nice to add it nevertheless...

sosna:
Close #427


javihern98 (approved):
Looks good, thanks! 😊

PR #421, opened 2025-10-06
Add SDMX-ML ProvisionAgreement readers and writers

jfrancma:
- Add support for Provision Agreement structure for XML readers and writers
- Added new function `__format_prov_agreement` to structure_aux_reader to read the content of a provision agreement
- Added new function `__write_prov_agreement` to structure_aux_writer to write the content of a provision agreement
- Added tests for all xml versions("2.1","3.0","3.1")  readers and writers

PR #440, opened 2025-10-17
Add support in get_datasets for ProvisionAgreement data references

jfrancma:
- Add support in get_datasets for ProvisionAgreement structure types
- Modify schema_generator to be able to generate ProvisionAgreement schemas

This PR also includes all the changes from #421 (Provision Agreement readers and writers) because they were necessary.

### DIFF
--- a/src/pysdmx/io/json/fusion/messages/category.py
+++ b/src/pysdmx/io/json/fusion/messages/category.py
@@ -1,7 +1,7 @@
 """Collection of Fusion-JSON schemas for categories and category schemes."""
 
 from collections import defaultdict
-from typing import Dict, Optional, Sequence
+from typing import Dict, Optional, Sequence, Tuple, Union
 
 from msgspec import Struct
 
@@ -11,6 +11,8 @@ from pysdmx.model import (
     Agency,
     Category,
     DataflowRef,
+    ItemReference,
+    Reference,
 )
 from pysdmx.model import (
     Categorisation as CT,
@@ -21,7 +23,7 @@ from pysdmx.model import (
 from pysdmx.model import (
     Dataflow as DF,
 )
-from pysdmx.util import find_by_urn
+from pysdmx.util import find_by_urn, parse_urn
 
 
 class FusionCategorisation(Struct, frozen=True, rename={"agency": "agencyId"}):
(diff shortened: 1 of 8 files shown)


jfrancma:
> Looks good, but please reduce the size of the test files to be as small as possible, especially the xml structures ones

Reduced the size of tests files

src/pysdmx/io/xml/__data_aux.py
```diff
@@ -92,9 +93,8 @@ def __get_elements_from_structure(structure: Dict[str, Any]) -> Any:
         structure_type = "ProvisionAgreement"
         tuple_ids = __get_ids_from_structure(structure[PROV_AGREEMENT])
     else:
-        raise NotImplemented(
-            "Unsupported", "ProvisionAgrement not implemented"
-        )
+        structure_type = "ProvisionAgreement"
```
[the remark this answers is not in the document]
jfrancma: The section above is for handling ProvisionAgreement, but this one is for handling ProvisionAgrement, which is a typo in the XML 2.1 schema. It was fixed in version 3.0.

src/pysdmx/model/message.py
```diff
@@ -235,26 +235,66 @@ def get_metadataflows(self) -> List[Metadataflow]:
         return self.__get_elements(Metadataflow)
 
     def get_organisation_scheme(self, short_urn: str) -> AgencyScheme:
-        """Returns a specific OrganisationScheme."""
-        return self.__get_single_structure(AgencyScheme, short_urn)
+        """Returns a specific AgencyScheme."""
+        obj = self.__get_single_structure(AgencyScheme, short_urn)
+        if not isinstance(obj, AgencyScheme):
```
sosna: Why do we need these additional checks in every function? The `short_urn` includes the artefact type and is generated by us based on the artefact type... I guess we can trust what we ourselves generated, no 😁? 

```python
    @property
    def short_urn(self) -> str:
        """Returns the short URN for the artefact.

        A short URN follows the syntax: Type=Agency:Id(Version). For example:
        Codelist=SDMX:CL_FREQ(1.0)

        Returns:
            The short URN for the artefact.
        """
        agency = (
            self.agency.id if isinstance(self.agency, Agency) else self.agency
        )
        return f"{self.__class__.__name__}={agency}:{self.id}({self.version})"
```

sosna (changes requested):

src/pysdmx/model/message.py
javihern98: I finally used typing as Any for __get_single_structure and added comments detailing the expected output. As they are internal functions later called by the publicly exposed functions and it is the best solution I could fine, I suggest to leave it as my latest push.

I tried everything AI or internet could provide me but it was overcomplicating things for an internal function, plus I prefer to add an Any type here (in an internal function) rather than add mypy ignore errors for this. Stuff I tried includes Generic typing, TypingVars and generating a list of the inherited classes for typing, but nothing works.

Not my desired solution but I can live with it 😊

sosna (approved):
That is a good solution, @javihern98, thanks 😊!

src/pysdmx/model/message.py
javihern98: Decided in the meeting this solution is correct, so resolving it

PR #445, opened 2025-10-30
Support multiple receivers in the header of SDMX messages

### LINKED ISSUE #402: Receivers in pysdmx Header
The `pysdmx.io.model.message.Header` class has a property called `receiver`, of type `Optional[Organisation]`. However, SDMX messages allow more than one receiver.

sosna:
Close #402


javihern98:
@sosna Just realised we did not check the SDMX-ML readers, I am restoring the branch as they seem to be failing with this change if more than one receiver is sent. Shall we do it in the same branch or generate a new issue for this? It fails in any SDMX-ML file with two receivers in the Header

PR #455, opened 2025-11-07
Add Python 3.13 to the CI pipeline

sosna:


javihern98:
@sosna I added some type ignores on mypy to remove the override error, still I cannot find a better solution than this and I cannot ignore these errors only for python 3.13. In any case, given the class hierarchy structure for SDMX, I believe we should always disable this override error (the signature of internal method of classes of a Codelist and an ItemScheme, for example, is different by design as the Codelist has more arguments.

Issue #460, opened 2025-11-14
ESTAT SDMX-CSV files show wrong dialect on sniffing


javihern98:
# Overview

When downloading some files from Eurostat SDMX API, the dialect is detected as | instead of commas. This may be caused by an internal bug in the Sniffer from python csv.

URLs:
- SDMX-CSV 1.0 (labels only): https://ec.europa.eu/eurostat/api/dissemination/sdmx/3.0/data/dataflow/ESTAT/ei_bpm6ca_q$defaultview/1.0?compress=false&format=csvdata&formatVersion=1.0&lang=en&labels=label_only
- SDMX-CSV 2.0 (labels and name): https://meaningfuldata.slack.com/archives/D0214NG5SLW/p1763118942968989

Files from the URLs:

[estat-1_0.csv](https://github.com/user-attachments/files/23546367/estat-1_0.csv)
[estat-2_0.csv](https://github.com/user-attachments/files/23546368/estat-2_0.csv)

Screenshot from debugger:

<img width="1837" height="320" alt="Image" src="https://github.com/user-attachments/assets/d969fe09-5593-4fe1-8191-0ab727164bfb" />

We can see the file is separated by commas but the delimiter is not properly set, therefore we should relax the conditions on the control_csv_format or change the sniffer used.

ruizmaa:
The failure comes from our [`__check_csv`](https://github.com/bis-med-it/pysdmx/blob/develop/src/pysdmx/io/input_processor.py#L30-L45) function, which uses `csv.Sniffer().sniff` on a sample limited to 2048 bytes:
```python
def __check_csv(input_str: str) -> bool:
    try:
        max_length = min(2048, len(input_str))
        dialect = csv.Sniffer().sniff(input_str[:max_length])
        control_csv_format = (
            dialect.delimiter == "," and dialect.quotechar == '"'
        )
        # Check we can access the data and it is not empty
        if (
            len(input_str.splitlines()) > 1
            or input_str.splitlines()[0].count(",") > 1
        ) and control_csv_format:
            return True
    except Exception:
        return False
    return False
```

For the files attached on this issue, when using a smaller sample (2048 bytes), `csv.Sniffer` missdetechts the delimiter (`I`, `f` or `P` instead of `,`)

But when the sample is larger it detects `,` as the delimiter correctly (from 4096 bytes).

Please find attached the script i used to generate the results and the .csv files used:

[...]

PR #465, opened 2025-11-19
Fix wrong dialect on sniffing

ruizmaa:
**IMPORTANT:** Read issue's #460 comments 

This PR fixes unreliable CSV detection caused by the previous 2048-byte sample used by `csv.Sniffer`.

`__check_csv` has been updated to use the first complete line as the `Sniffer` sample instead of a fixed byte buffer. This prevents delimiter misdetection caused by truncated samples and makes CSV format inference robust for large SDMX-CSV

Fixed issue #460


sosna (approved):
Nice, thanks

PR #471, opened 2025-11-27
Find items in item schemes by name or description

sosna:
The purpose of this PR is to support finding items in an item scheme by name and/or description. 

By default, a partial and case-insensitive search is performed on both fields, but it is possible to tweak this using regular expressions and by defining the fields to be considered for the search.


javihern98 (approved):
Looks like a really nice addition, thanks!😊

PR #485, opened 2025-12-18
Implement support for maps in SDMX-ML 

### LINKED ISSUE #458: Expand pysdmx.io.read_sdmx to read certain SDMX objects with SDMX-ML
I have tried using pysdmx.io.read_sdmx to read DSDs, Dataflows, StructureSets and RepresenatationMaps. My goal would be to replace custom SDMX object representations in the mappings_utility package with the ones implements in pysdmx.models.

However the read_sdmx module seems to only read the DSDs and Dataflows at the moment, but ignore other arterfacts. Would it be possible to have a more complete reader? Or is there another recommended mechanism to read structure sets and representation maps?

ruizmaa:
Fixes #458

This PR expands the capabilities of the SDMX-ML reader and writer to support Mapping structures. Following the [SDMX mapping schema](https://github.com/Meaningful-Data/sdmx-schemas/blob/main/src/sdmxschemas/xml/sdmx31/SDMXStructureStructureMappings.xsd)

## Supported mapping types

Previously, read_sdmx ignored these artifacts. With this change, the following objects can now be parsed from XML structure messages:

- `StructureMap`
- `ComponentMap`
  - 1–1 to `ComponentMap`
  - n–m and 1-n/n-1 to `MultiComponentMap`
  - `ImplicitComponentMap`
- `FixedValueMap`
- `RepresentationMap`
- `ValueMap`
- `MultiValueMap`
- `DatePatternMap`
  - fixed and variable pattern types

### Reader (`io/xml/__structure_aux_reader.py`)

- Extended `StructureParser` to collect "top-level" mapping objects:
  - `StructureMap`
  - `ComponentMap`
  - `FixedValueMap`
  - `RepresentationMap`
- "Children" mapping objects (non-maintainable) are parsed and attached to their parent mapping object.
  - `ValueMap`
  - `MultiValueMap`
  - `MultiComponentMap`
  - `ImplicitComponentMap`
  - `DatePatternMap`
- New `__format_maps` helper: a recursive parser to handle map objects which normalizes the XML tags names (e.g., `SourceCodelist`, `SourceValue`, `Source`, `RepresentationMap`) to the model attributes (`source`, `target`, `values`) and infers the correct mapping types (`ComponentMap`, `MultiComponentMap`, `ImplicitComponentMap`, ...)

### Writer (`io/xml/__structure_aux_writer.py`)

- Representation Maps: added `__write_representation_map` to serialize  `RepresentationMap` objects
  - Writes `SourceCodelist` and `TargetCodelist` references.
  - Serializes mapping child objects with:
    - `__write_value_map`: `ValueMap` (1-1) and `MultiValueMap` (n–m and 1-n/n-1)
- Structure Maps: added `__write_structure_map` to serialize `StructureMap` objects
  - Writes `SourceCodelist` and `TargetCodelist` references
  - Serializes mapping child objects with:
    - `__write_date_pattern_map` for `DatePatternMap` (fixed and variable patterns)
    - `__write_component_map` for 1–1 `ComponentMap`
    - `__write_multi_component_map` for n–m / 1–n / n–1 `MultiComponentMap`
    - `__write_implicit_component_map` for `ImplicitComponentMap`
    - `__write_fixed_value_map` for `FixedValueMap`

#### Auxiliary changes (`src/pysdmx/io/xml/__write_aux.py`)

- `__write_scheme` supports `StructureMap` and `RepresentationMap`
- Added `StructureMaps` and `RepresentationMaps` to `STR_DICT_TYPE_LIST_21` and `STR_DICT_TYPE_LIST_30`

> Only `StructureMap` and `RepresentationMap` are included here, as they are the only mapping artefacts defined as `MaintainableType` in the SDMX schema.

## Tests

- Added a sample file `samples/maps.xml` with all the mapping structures
- Reader tests (`test_general_reader.py`): verifies parsing every map and checks include object types, sources/targets/values, optional fields, and iteration/length behavior.
- Writer tests (`test_general_writer.py`): verifies that the maps are correctly serialized and that a round-trip read/write preserves the original structures.


javihern98 (commented):
Missing writers for these objects in SDMX-ML, we need to implement them as well

sosna (commented):
Thanks. The code reads very nicely 😊.

However, I'm wondering whether the added support for maps is not partial only. In the supplied test file, I cannot see the following types of very common mappings:

- Implicit mappings
- Date pattern maps
- Components maps with a 1-to-n, n-to-1 or n-to-n relationship.

These are all supported by the SDMX-JSON serializers / deserializers and I guess we'll want to do the same for the SDMX-ML ones?

sosna:
Thanks for the clarification, @javihern98. I had missed that 😊. I'd say that mapping is probably useless with implicit mapping (it's a very common use case when mapping DSDs). The other two are very popular too. So much that I'm wondering if it's worth proceeding without them.

Of course, we can add support in two steps as well, one now, one later, if this has your preference. But, in my understanding, usefulness will be limited without full support.

On the other hand, SDMX-JSON should be the default for structural metadata since SDMX 3.0, and our SDMX-JSON deserializers support all of these.

In short, happy with whatever option has your preference 😊

sosna (approved):
Great work, thanks!

PR #527, opened 2026-02-12
Fix issue with SDMX-JSON metadata messages not being properly recognized

### LINKED ISSUE #525: Bug when attempting to read a reference metadata report
The following fails (on our internal network):

```python
from pysdmx.api.qb import ApiVersion, RefMetaByMetadataflowQuery, RestService


query = RefMetaByMetadataflowQuery("BIS.MEDIT", "MEDAL_DF_CNF", "1.0")
service = "https://mmsu.bisinfo.org/sdmx/v2"
url = f"{service}{query.get_url(ApiVersion.V2_0_0, omit_defaults=True)}"

msg = read_sdmx(url, validate=False, pem="bis_certs.pem")
```

This is because the function trying to infer the type of JSON message wrongly reports it as being a structure message.

sosna:
This small PR closes the bug described in issue #525. 

Whitespaces in the JSON files have been removed and a link object has been added to one of them (both changes were necessary to reproduce the issue), but else, they are unchanged. 

The problem was that the links contained a reference to a structure and as structure schema strings were checked before reference metadata ones, the wrong type was inferred.


javihern98 (dismissed):
I understand it works now with both whitespaces and without them. Therefore it looks better now, but I would try to keep the tests with whitespaces and make a json.dumps with no indent here. Just my two cents😊

javihern98 (approved):
Looks much better now, thanks!😊

sosna:
My pleasure and thanks for approving 😊

PR #528, opened 2026-02-12
Unmerge attributes with multiple values when writing to SDMX-JSON

### LINKED ISSUE #523: Bug when writing "consolidated" metadata attributes
A metadata report supports multiple occurrences of the same attribute. 

`pysdmx` groups these together (using the pysdmx.model.metadata.merge_attributes) when reading metadata reports.

However, it does not do the reverse when writing, i.e. it does not unpack the values that were consolidated when reading.

This leads to files containing values like this one (a list serialized as string):

```json
{
    "id":"DF_OWNERS",
    "value":"[ALIAS:urn:sdmx:org.sdmx.infomodel.structuremapping.RepresentationMap=BIS.TEST:ALIAS_2_USER(1.0), REF_AREA:urn:sdmx:org.sdmx.infomodel.structuremapping.RepresentationMap=BIS.TEST:REF_AREA_2_USER(1.0)]"
}
```

Instead of one attribute with two values, it should be two attributes with one value:

```json
{
    "id":"DF_OWNERS",
    "value":"ALIAS:urn:sdmx:org.sdmx.infomodel.structuremapping.RepresentationMap=BIS.TEST:ALIAS_2_USER(1.0)"
},
{
    "id":"DF_OWNERS",
    "value":"REF_AREA:urn:sdmx:org.sdmx.infomodel.structuremapping.RepresentationMap=BIS.TEST:REF_AREA_2_USER(1.0)"
}
```

In addition, we need to check how the consolidation works for reports that contain presentational attributes.

sosna:
In the same way that we merge multiple attributes with the same ID when reading SDMX-JSON metadata reports, we need to unmerge them, when writing SDMX-JSON reports.

Additional details are available in #523

### DIFF
--- a/src/pysdmx/io/json/sdmxjson2/messages/report.py
+++ b/src/pysdmx/io/json/sdmxjson2/messages/report.py
@@ -20,6 +20,7 @@ from pysdmx.model.metadata import (
     MetadataAttribute,
     MetadataReport,
     merge_attributes,
+    unmerge_attributes,
 )
 
 
@@ -105,7 +106,7 @@ class JsonMetadataReport(ItemSchemeType, frozen=True, omit_defaults=True):
                 "SDMX-JSON metadata reports must have a name",
                 {"metadata_report": report.id},
             )
-
+        attrs = unmerge_attributes(report.attributes)
         return JsonMetadataReport(
             agency=(
                 report.agency.id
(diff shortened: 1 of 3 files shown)


javihern98 (approved):
Looks good, thanks!😊

PR #529, opened 2026-02-13
Support observation-level attachments in measureless DSDs

### LINKED ISSUE #524: Bug when inferring attachment level of measureless DSDs
It is possible to have DSDs without measures and with attributes attached to the observation-level.

In that case, pysdmx fails to infer the attachment level because the property with the measure references is missing.

sosna:
It is possible to have DSDs without measures and with attributes attached to the observation-level.

In that case, pysdmx fails to infer the attachment level because the property with the measure references is missing.

The purpose of this PR is to address this issue.

Close #524


javihern98 (approved):
Looks good, thanks!😊

PR #532, opened 2026-02-19
Add SDMX-ML constraints support

### LINKED ISSUE #530: Add SDMX-ML constraints support
Data constraints (DataConstraint, ConstraintAttachment, CubeRegion, CubeKeyValue, CubeValue, KeySet, DataKey, DataKeyValue) are supported in the model classes and in SDMX-JSON (check PR #468 ), but they are not in SDMX-ML

Support should be implemented to both reader and writer.

ruizmaa:
Closes #530

This PR implements read/write support for `DataConstraint` structures in SDMX-ML (both SDMX 2.1 and SDMX 3.0 formats). The model classes (`DataConstraint`, `ConstraintAttachment`, `CubeRegion`, `CubeKeyValue`, `CubeValue`, `KeySet`, `DataKey`, `DataKeyValue`) were already supported in SDMX-JSON (PR #468).

## Changes

The changes have been made using the [SDMX XML schemas](https://github.com/Meaningful-Data/sdmx-schemas/tree/main/src/sdmxschemas/xml) as reference.

### Reader (`__structure_aux_reader.py`)

The reader now recognises `ContentConstraint` elements (2.1) and `DataConstraint` elements (3.0) and maps both to the existing `DataConstraint` model class. Both formats are registered in `STRUCTURES_MAPPING` to include them in the structure processing.

Parsing is done by private helper methods:

- `__parse_data_provider`: resolves the `DataProvider` reference, handling the SDMX 2.1 `<Ref>` element style and the SDMX 3.0 direct URN style.
- `__parse_references`: generic helper used to extract and normalise references to dataflows, data structures, and provision agreements into URN strings.
- `__format_constraint_attachment`: combines the above to produce a `ConstraintAttachment` object.
- `__format_cube_region`: walks the `CubeRegion` element and builds the nested `CubeRegion` / `CubeKeyValue` / `CubeValue` objects.
- `__format_key_set`: walks the `DataKeySet` element and builds the nested `KeySet` / `DataKey` / `DataKeyValue` objects.
- `__format_constraint`: top-level method called for each constraint element; it orchestrates the methods above and returns a dict ready to be passed to the `DataConstraint` constructor. Raises `NotImplementedError` if a parsed constraint has `role="Actual"`.

### Writer (`__structure_aux_writer.py`)

`DataConstraint` is written in the 2.1 type list (as `"Constraints"`) and in the 3.0 type list (as `"DataConstraints"`).

The serialisation is done by `__write_data_constraint`, which decides to emit a `<str:ContentConstraint>` (2.1) or a `<str:DataConstraint>` (3.0) element, and then delegates to a set of focused helpers:

- `__write_constraint_attachment`: writes the `<ConstraintAttachment>` and calls the appropriate reference helper for each attached artefact (`__write_data_provider`, `__write_data_structure`, `__write_dataflow`, `__write_provision_agreement`). Each of these handles the format difference: 3.0 writes a plain URN as text content, while 2.1 writes a `<Ref>` child element with individual attributes.
- `__write_cube_region`: serialises a `CubeRegion` with its `KeyValue`/`Value` children. The namespace prefix differs per format (`str:` in 3.0, `com:` in 2.1).
- `__write_key_set`: serialises a `DataKeySet` with its `Key`/`KeyValue`/`Value` children, following the same namespace convention.

### Fix: Group URN generation in XML writer (`__structure_aux_writer.py`)

When `Group` elements were created from JSON sources (which don't include URNs), the XML writer would output `urn=None`, producing invalid XML. The `__write_groups` function now generates a proper URN from the parent DSD context (`agency`, `id`, `version`) when the group's URN is not set, using the format `urn:sdmx:org.sdmx.infomodel.datastructure.GroupDimensionDescriptor={agency}:{dsd_id}({version}).{group_id}`.

### Fix: Optional `constraintAttachment` in SDMX-JSON (`constraint.py`)

The `JsonDataConstraint.to_model()` method now handles `None` `constraintAttachment` gracefully instead of raising an `AttributeError`.

## Tests

New sample XML files and test cases are added:

- Constraint with `CubeRegion` attached to a `Dataflow`
- Constraint with `DataKeySet`
- Constraint attached to a `DataStructure`
- Constraint attached to a `DataProvider`
- Constraint attached to a `ProvisionAgreement`
- Constraint without `ConstraintAttachment`
- Group without URN written to SDMX-ML 2.1, 3.0, and 3.1 (validates generated URN and round-trips through reader)
- Constraint deserialization from SDMX-JSON without `constraintAttachment`

All writer tests validate the output by round-tripping through the reader (`read`/`validate=True`) before asserting against the expected XML sample.

## Out of Scope

- Added writer for `ProvisionAgreement` and reader/writer for `DataProvider`

### DIFF
--- a/src/pysdmx/io/xml/__structure_aux_writer.py
+++ b/src/pysdmx/io/xml/__structure_aux_writer.py
@@ -1150,12 +1150,24 @@ def __write_structure_map(
     return outfile
 
 
-def __write_data_provider(data_provider: str, indent: str) -> str:
+def __write_data_provider(
+    data_provider: str, indent: str, references_30: bool = False
+) -> str:
     """Writes a DataProvider reference to the XML file."""
-    outfile = f"{indent}<{ABBR_STR}:DataProvider>"
-    outfile += f"{add_indent(add_indent(indent))}<{REF} "
-    outfile += f"{ID}={data_provider!r}/>"
-    outfile += f"{indent}</{ABBR_STR}:DataProvider>"
+    outfile = f"{add_indent(indent)}<{ABBR_STR}:DataProvider>"
+    if references_30:
+        # SDMX 3.0: URN
+        outfile += data_provider
+    else:
+        # SDMX 2.1: Ref with attributes
+        outfile += f"{add_indent(add_indent(indent))}<{REF} "
+        ref = parse_urn(data_provider)
+        outfile += f"{AGENCY_ID}={ref.agency!r} "
+        outfile += f"{PAR_ID}={ref.id!r} "
+        outfile += f"{PAR_VER}={ref.version!r} "
+        outfile += f"{ID}={ref.item_id!r} "
+        outfile += f"{CLASS}={DATA_PROV!r}/>"
+    outfile += f"</{ABBR_STR}:DataProvider>"
     return outfile
 
 
@@ -1226,7 +1238,9 @@ def __write_constraint_attachment(
 
     # DataProvider
     if attachment.data_provider:
-        outfile += __write_data_provider(attachment.data_provider, indent)
+        outfile += __write_data_provider(
+            attachment.data_provider, indent, references_30
+        )
 
     # DataStructures
     if attachment.data_structures:
--- a/src/pysdmx/io/xml/__structure_aux_writer.py
+++ b/src/pysdmx/io/xml/__structure_aux_writer.py
@@ -16,12 +16,17 @@ from pysdmx.io.xml.__tokens import (
     CLASS,
     COMPONENT_MAP,
     CON,
+    CON_CONS,
     CON_ID,
     CONDITIONAL,
+    CONS_ATT,
     CORE_REP,
     CS,
+    CUBE_REGION,
     CUSTOM_TYPE,
     CUSTOM_TYPE_SCHEME,
+    DATA_CONS,
+    DATA_KEY_SET,
     DATA_PROV,
     DEPARTMENT,
     DFW,
@@ -37,6 +42,10 @@ from pysdmx.io.xml.__tokens import (
     GROUP_DIM,
     GROUPS_LOW,
     ID,
+    INCLUDE,
+    INCLUDED,
+    KEY,
+    KEY_VALUE,
     LOCAL_REP,
     MANDATORY,
     MANDATORY_LOW,
@@ -71,6 +80,7 @@ from pysdmx.io.xml.__tokens import (
     URN,
     URN_LOW,
     USAGE,
+    VALUE,
     VALUE_ITEM,
     VALUE_LIST,
     VALUE_LIST_LOW,
@@ -95,14 +105,18 @@ from pysdmx.model import (
     ComponentMap,
     Concept,
     ConceptScheme,
+    ConstraintAttachment,
+    CubeRegion,
     CustomType,
     CustomTypeScheme,
+    DataConstraint,
     DataType,
     DatePatternMap,
     Facets,
     FixedValueMap,
     Hierarchy,
     ImplicitComponentMap,
+    KeySet,
     MultiComponentMap,
     MultiValueMap,
     NamePersonalisation,
@@ -185,6 +199,7 @@ STR_DICT_TYPE_LIST_21 = {
     ConceptScheme: "Concepts",
     DataStructureDefinition: "DataStructures",
     Dataflow: "Dataflows",
+    DataConstraint: "Constraints",
     RepresentationMap: "RepresentationMaps",
     StructureMap: "StructureMaps",
     DatePatternMap: "DatePatternMaps",
@@ -204,6 +219,7 @@ STR_DICT_TYPE_LIST_30 = {
     ConceptScheme: "ConceptSchemes",
     DataStructureDefinition: "DataStructures",
     Dataflow: "Dataflows",
+    DataConstraint: "DataConstraints",
     RepresentationMap: "RepresentationMaps",
     StructureMap: "StructureMaps",
     DatePatternMap: "DatePatternMaps",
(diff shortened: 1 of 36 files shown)


src/pysdmx/io/xml/__structure_aux_writer.py
```diff
@@ -1134,6 +1150,238 @@ def __write_structure_map(
...
+    data = __write_maintainable(constraint, indent, references_30)
+
+    # Add role attribute for SDMX 3.0 (required)
+
+    # TODO: add 'role' field to 'DataConstraint'
+    # to know which role to use (Allowed | Actual)
+    # In SDMX-JSON role is ignored and
+    # was deleted from JsonDataConstraint (PR #468)
+    if references_30:
+        data["Attributes"] += ' role="Allowed"'
```
ruizmaa: Not sure about what should i do with this. The [.xsd](https://github.com/Meaningful-Data/sdmx-schemas/blob/main/src/sdmxschemas/xml/sdmx30/SDMXStructureConstraint.xsd) marks role as required:

```xml
<xs:attribute name="role" type="ConstraintRoleType" use="required">
	<xs:annotation>
		<xs:documentation>The role attribute indicates whether this constraint states what data is actually present for the constraint attachment, or if it defines what content is allowed.</xs:documentation>
	</xs:annotation>
</xs:attribute>
```

```xml
<xs:simpleType name="ConstraintRoleType">
	<xs:annotation>
		<xs:documentation>ConstraintRoleType defines a list of roles for a content constraint. A constraint can state which data is present or which content is allowed for the constraint attachment.</xs:documentation>
	</xs:annotation>
	<xs:restriction base="xs:string">
		<xs:enumeration value="Allowed">
			<xs:annotation>
				<xs:documentation>The constraint contains the allowed values for attachable object.</xs:documentation>
			</xs:annotation>
		</xs:enumeration>
		<xs:enumeration value="Actual">
			<xs:annotation>
				<xs:documentation>The constraints contains the actual data present for the attachable object.</xs:documentation>
			</xs:annotation>
		</xs:enumeration>
	</xs:restriction>
</xs:simpleType>
```

But the [class](https://github.com/bis-med-it/pysdmx/blob/develop/src/pysdmx/model/constraint.py#L64) does not have the attribute and the [json class used to have it but was deleted](https://github.com/bis-med-it/pysdmx/pull/468/changes#diff-bc1236bb7e839b49b15ffefeeaacd1b09f9cf73032ac660f6fc8bf327fc8c11dL51)

So im confused with this

javihern98 (commented):

src/pysdmx/io/xml/__structure_aux_writer.py
javihern98: @sosna Could you have a look at this? What do you think?

sosna: @stratosn will correct me if I'm wrong, but I believe this was a cleanup done during the 3.1 standard release (cf. point 3.2 on page 6 of https://sdmx.org/wp-content/uploads/SDMX_3-1-0_Major_Changes_FINAL-1_0.pdf).

As you can see in SDMX-JSON 2.1.0, role is no longer present. It was present in SDMX 2.0.0, but I don't believe it was mandatory.

I believe it is missing from SDMX-ML 3.1 too, but it is still there in earlier versions of SDMX-ML.

To summarize:

- In SDMX, only "allowed" data constraints are supposed to be maintainable. "available" data constraints aren't, as they are supposed to be generated on-the-fly.
- In pysdmx, we support only the maintainable ones, i.e. "allowed" data constraints.
- This ought to be good enough, IMO, but if you believe this is wrong, we have two options:
    - Adding back a role property, which in principle would be wrong as available constraints are not maintainable artefacts.
    -  Adding a second type of constraints, i.e. AvailabilityConstraint, not inheriting from Maintainable. This makes the model slightly more complex though, so I would refrain from doing that, unless we really have to.

So, I'd vote for ignoring the old "available" constraints, as they have been deprecated, for good reasons. What do you think?

stratosn: Indeed, there was a clean-up in SDMX 3.1. 
I am only wondering whether we should explicitly make the distinction between the two types for SDMX 3.0 messages and thus, respond "appropriately" in case the type is "available".

By "appropriately" I mean either "Not Implemented" when someone queries an endpoint that returns such a  Constraint, or in general on parsing.
`pysdmx` should not be generating such messages, ie of "available" type.

ruizmaa: Thanks for the clarification @sosna and @stratosn! I afree with both of you

The writer already hardcodes `role="Allowed"` in the output, so no changes needed there.
I'll just update the reader to raise a `NotImplementedError` when it encounters `role="Actual"`

javihern98: @sosna @ruizmaa @stratosn Is a NotImplementedError the best approach here? Shouldn´t we just ignore it?

sosna: @javihern98, the NotImplemented error sounds OK to me, as we indeed don't have a object in the model (yet?) for this type of constraints. I'm not sure we should silently ignore this (I'm a big fan of PEP-20 😁).

javihern98: Ok, keeping the NotImplementedError then :blush:

javihern98:
@sosna This is still pending for your review to publish 1.13.0 :blush:

sosna:
@javihern98, I know, sorry 😞. I have been bogged down with internal tasks but I'm about to get free soon 😊

javihern98:
@sosna No problem at all! 😊

sosna:
I have run the test that I mentioned previously (reading a SDMX-JSON structure file with constraints and their descendants, writing to SDMX-ML, reading the generated SDMX-ML, and comparing the objects), and it looks good 😊.

There were only two minor issues, i.e. the comparison failed for two reasons:

- The XML artefacts had the `isValid` property set to True, while the JSON ones didn't. I believe the JSON one is correct, or at least it seemed aligned with the SDMX 3.0 spec and the new semver rules. But if you disagree, happy to discuss!
- The Agency Scheme in JSON didn't return the name and id, because both are fixed (set to AGENCIES), while the XML had the name and id.

```
json = 
AgencyScheme(agency='SDMX', items=[Agency(id='BIS', name='Bank for International Settlements', description='Bank for International Settlements', contacts=[Contact(name='Support', role='Support', emails=['statistics@bis.org'])])], is_partial=True)

xml =
AgencyScheme(id='AGENCIES', name='AGENCIES', is_final=True, agency='SDMX', items=[Agency(id='BIS', name='Bank for International Settlements', description='Bank for International Settlements', contacts=[Contact(name='Support', role='Support', emails=['statistics@bis.org'])])], is_partial=True)
```

Both are unrelated to adding support for constraints in SDMX-ML so, if you agree with this, I can approve the PR. However, it would be good to later work on the 2 issues mentioned above. @ruizmaa, @javihern98, what do you think?

javihern98:
Hi @sosna, I agree to finish the approval on this and raise another issues to investigate the isValid property and the id and name. I will generate both issues :blush:

sosna (approved):
Looks good, thanks 😊!

PR #533, opened 2026-02-19
Add support for exclusive keysets

### LINKED ISSUE #526: Add support for exclusive keysets
Support for keysets in schemas is currently limited to inclusive keysets. We should support exclusive keysets too.

sosna:
This PR adds support for exclusive keysets to SDMX-JSON and Fusion-JSON.

While we initially thought this was more of an exotic use case, there is actually a need for this.


javihern98 (approved):
Looks good, thanks! 😊

ruizmaa (approved):
All good from my side! Thank you

PR #538, opened 2026-02-26
fix: handle DSD instances in schema_generator (#534)

### LINKED ISSUE #534: Issue with get_datasets
I am trying to read the attached data file and the attached structure file, using `get_datasets`. The code is as follows:

```python
from pysdmx.io import get_datasets, write_sdmx


data = get_datasets("data.in.xml", "structure.json", validate=False)
write_sdmx(data, Format.DATA_SDMX_CSV_2_1_0, "data.out.csv")
```

This is the stacktrace I get:

```
Traceback (most recent call last):
  File "c:\Users\xa003754\Dev\pysdmx\convert.py", line 5, in <module>
    data = get_datasets("data.in.xml", "structure.json", validate=False)
  File "C:\Users\xa003754\Dev\pysdmx\src\pysdmx\io\reader.py", line 288, in get_datasets
    __assign_structure_to_dataset(data_msg.data, structure_msg)
  File "C:\Users\xa003754\Dev\pysdmx\src\pysdmx\io\reader.py", line 225, in __assign_structure_to_dataset
    dataset.structure = schema_generator(structure_msg, dataset_ref)
  File "C:\Users\xa003754\Dev\pysdmx\src\pysdmx\util\_model_utils.py", line 30, in schema_generator
    dsd_ref = parse_urn(dataflow.structure)  # type: ignore[arg-type]
  File "C:\Users\xa003754\Dev\pysdmx\src\pysdmx\util\__init__.py", line 22, in parse_urn
    return parse_maintainable_urn(urn)
  File "C:\Users\xa003754\Dev\pysdmx\src\pysdmx\util\__init__.py", line 40, in parse_maintainable_urn
    m = re.match(maintainable_urn_pattern, urn)
  File "C:\Program Files\Python39\lib\re.py", line 191, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or bytes-like object
```

This seems to be because of line 30 in _model_utils.py, where we try and parse the DSD urn as if it was a string. But, in this particular case, we already have a Data Structure instance there.

The following code snippet fixes the issue but I don't know  whether it's the right thing to do:

```python
if isinstance(dataflow.structure, str):
    dsd_ref = parse_urn(dataflow.structure)  # type: ignore[arg-type]
    try:
        dsd = message.get_data_structure_definition(str(dsd_ref))
    except NotFound:
        raise Invalid(
            f"Not found referenced DataStructure {dsd_ref}"
            f"from Dataflow {dataset_ref}. "
            f"Please send the structures message using "
            f"references=children to include the DataStructureDefinition.",
        ) from None
else:
    dsd = dataflow.structure
```

[data.in.xml](https://github.com/user-attachments/files/25416861/data.in.xml)

[structure.json](https://github.com/user-attachments/files/25416864/structure.json)

javihern98:
## Summary
- Fix `TypeError` when `Dataflow.structure` is already a `DataStructureDefinition` instance instead of a URN string, which caused `parse_urn` to fail
- Add `isinstance` check for both `dataflow` and `provisionagreement` branches
- Extract `_resolve_dsd` and `_build_schema` helpers to eliminate duplicated DSD resolution and Schema construction logic
- Remove `# noqa: C901` complexity override

### DIFF
--- a/src/pysdmx/util/_model_utils.py
+++ b/src/pysdmx/util/_model_utils.py
@@ -67,8 +67,7 @@ def schema_generator(
             dsd = message.get_data_structure_definition(str(dataset_ref))
         except NotFound:
             raise Invalid(
-                f"Missing DataStructure {dataset_ref} "
-                f"in structures message.",
+                f"Missing DataStructure {dataset_ref} in structures message.",
             ) from None
         return dsd.to_schema()
     elif context == "dataflow":
@@ -100,8 +99,7 @@ def schema_generator(
             dataflow = message.get_dataflow(str(dfw_ref))
         except NotFound:
             raise Invalid(
-                f"Missing Dataflow in {dataset_ref} "
-                f"structures message.",
+                f"Missing Dataflow in {dataset_ref} structures message.",
             ) from None
         dsd = _resolve_dsd(
             dataflow, message, dataset_ref, "Provision Agreement"
--- a/src/pysdmx/util/_model_utils.py
+++ b/src/pysdmx/util/_model_utils.py
@@ -1,11 +1,65 @@
 from pysdmx.errors import Invalid, NotFound
 from pysdmx.model import Reference
-from pysdmx.model.dataflow import Schema
+from pysdmx.model.dataflow import (
+    Dataflow,
+    DataStructureDefinition,
+    Schema,
+)
 from pysdmx.model.message import Message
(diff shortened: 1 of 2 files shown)


sosna (approved):
Looks good, thanks 😊

sosna:
I linked the PR with issue #534, which I believe is the correct one for this PR 😊

PR #552, opened 2026-03-12
Remove incorrect is_final inference from XML 3.0 reader

### LINKED ISSUE #544: Issues with IsFinal as discrepancy between SDMX-JSON and SDMX-ML
## Description

When reading the same SDMX structure from SDMX-JSON and SDMX-ML 3.0, the `is_final` property on maintainable artefacts differs:

- **SDMX-ML 3.0 reader**: Infers `is_final=True` when the version does not contain a `-` extension (via `__format_is_final_30()`), even when the XML source does not include an explicit `isFinal` attribute.
- **SDMX-JSON reader**: Does not set `is_final`, so it defaults to `False`.

This means that reading the same logical artefact from two formats produces objects that are not equal.

## Root Cause

The XML reader in `__structure_aux_reader.py` applies an SDMX 3.0 semver rule in `__format_is_final_30()`: if the version has no `-` extension, `is_final` is forcibly set to `True`. This inference is incorrect — `is_final` should only be `True` when explicitly set in the source data.

Per [this comment](https://github.com/bis-med-it/pysdmx/pull/532#issuecomment-4005010822) from @sosna, the JSON reader behavior (defaulting to `False`) is correct and aligned with the SDMX 3.0 spec.

## Fix

Remove the `__format_is_final_30()` method and its calls from the XML reader. The explicit `isFinal` attribute parsing (for when the attribute is actually present in the XML) is preserved.

## Expected behavior

Reading the same logical structure from SDMX-JSON and SDMX-ML 3.0 should produce objects with the same `is_final` value. The `is_final` property should only be `True` when explicitly set in the source, not inferred from the version string.

_Originally posted by @sosna in https://github.com/bis-med-it/pysdmx/pull/532#issuecomment-4005010822_

javihern98:
## Summary
- Removed `__format_is_final_30()` from the SDMX-ML 3.0 reader, which incorrectly inferred `is_final=True` from version strings without a `-` extension
- `is_final` is now only set to `True` when the `isFinal` attribute is explicitly present in the XML source
- Fixes the discrepancy where reading the same artefact from SDMX-JSON and SDMX-ML 3.0 produced different `is_final` values

Closes #544

### DIFF
--- a/src/pysdmx/io/json/sdmxjson2/messages/agency.py
+++ b/src/pysdmx/io/json/sdmxjson2/messages/agency.py
@@ -11,6 +11,7 @@ from pysdmx.io.json.sdmxjson2.messages.core import (
 )
 from pysdmx.io.json.sdmxjson2.messages.dataflow import JsonDataflow
 from pysdmx.model import Agency, AgencyScheme, DataflowRef
+from pysdmx.util import is_final
 
 
 def _sanitize_agency(agency: Agency, is_sdmx_scheme: bool) -> Agency:
@@ -56,6 +57,7 @@ class JsonAgencyScheme(ItemSchemeType, frozen=True, omit_defaults=True):
             items=agencies,
             annotations=tuple([a.to_model() for a in self.annotations]),
             is_external_reference=self.isExternalReference,
+            is_final=is_final(self.version),
             is_partial=self.isPartial,
             valid_from=self.validFrom,
             valid_to=self.validTo,
--- a/src/pysdmx/io/json/sdmxjson2/messages/category.py
+++ b/src/pysdmx/io/json/sdmxjson2/messages/category.py
@@ -23,7 +23,7 @@ from pysdmx.model import (
     ItemReference,
     Reference,
 )
-from pysdmx.util import find_by_urn, parse_urn
+from pysdmx.util import find_by_urn, is_final, parse_urn
 
 
 class JsonCategorisation(
@@ -48,6 +48,7 @@ class JsonCategorisation(
             name=self.name,
             description=self.description,
             is_external_reference=self.isExternalReference,
+            is_final=is_final(self.version),
             valid_from=self.validFrom,
             valid_to=self.validTo,
             annotations=[a.to_model() for a in self.annotations],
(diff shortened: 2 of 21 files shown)


sosna:
Hi @javihern98. I have added 2 tests which I believe will prove that this item needs further work. I suspect SDMX-JSON deserializers will be wrong too. Let me explain 😊.

Before 3.0, we needed the `isFinal` flag to know whether an artefact could be considered final (i.e. not suspected to change anymore).

Starting with SDMX 3.0, which introduced support for semantic versioning, this is no longer necessary. Artefacts that use semantic versioning can have their final flag inferred from their version. For example, `1.0.0` is final, while `1.0.0-beta`, isn't. In short, any string appearing after the patch number identify the artefact as non final. Furthermore, artefacts that do not use semantic versioning (e.g. their version has only two digits) are all considered as non final. This is why the isFinal flag is gone from the SDMX 3.0 formats.

Therefore, we could do something like the following:
- Add a utility function (under `pysdmx.util`) to infer, from its version, whether an artefact is final. Something like:
```python
def is_final(version: str) -> bool:
    pass
``` 

- The deserializers of formats before SDMX 3.0 would not use that function, i.e. they would infer the final flag from appropriate attribute, like we do now.
- The deserializers of format from SDMX 3.0 and above would use that function to infer the value of the final flag from the version string.

Once we do that, the failing test I added should pass. What do you think?

@stratosn, let me know in case you think something is wrong in the above.

javihern98:
Hi @sosna, I like the approach, I could include the changes in SDMX-JSON in this branch as well.

Just a quick comment, the [SDMX 3.1 Technical Notes](https://sdmx.org/wp-content/uploads/SDMX_3-1-0_SECTION_6_FINAL.pdf) says:

image

Should this be considered as well as is_final=False?

javihern98:
> Should versions starting with 0. this be considered as well as is_final=False?

I implemented it like this :blush:

sosna:
And you were correct! Thanks for spotting this, I had forgotten it already 😊.

sosna (approved):
Really nice work, Javier, thanks! I think we have something solid now 😊

PR #554, opened 2026-03-12
Fix PandasDataset typing for get_datasets

### LINKED ISSUE #543: (Pandas)Dataset and type hints / autocompletion
Message.get_datasets() returns a list of Dataset in its signature.
The actual implementation returns a PandasDataset.

The problem with python is that it is not possible to cast the result to PandasDataset, and therefore we do not have access to its functoins during the development phase. the type hints fail and there is no autocompletion for pandas.

The only way I can think of is to do this to force it:
Solution 1: 
message = read_sdmx(dst_file)
dataset: PandasDataset = message.get_datasets()[0]

Solutin 2:
message = read_sdmx(dst_file)
df = pd.DataFrame(message.get_datasets()[0].data)

In both cases, my dev environment does not like it, and without the documentation or the debug mode, we do not know what we expect.

Would it be please possible to do one of the following:
- Add the method Message.get_pandas_datasets() or,
- Add the method Dataset.to_PandasDataset() or
- something else

javihern98:
## Summary
- Change return types of `Message.get_datasets()`, `Message.get_dataset()`, and `io.get_datasets()` from `Dataset` to `PandasDataset`
- Add `@overload` decorators to `io.get_datasets()` dispatched on the `structure` parameter
- Users now get proper IDE autocompletion for `PandasDataset.data` (`pd.DataFrame`)

Closes #543

## Test plan
- [x] All 4555 existing tests pass
- [x] 100% branch coverage maintained
- [x] mypy strict mode passes
- [x] ruff format and ruff check pass

### DIFF
--- a/src/pysdmx/io/reader.py
+++ b/src/pysdmx/io/reader.py
@@ -8,6 +8,7 @@ from typing import (
     Optional,
     Sequence,
     Union,
+    cast,
     overload,
 )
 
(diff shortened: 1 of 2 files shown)


sosna (approved):
Looks good, thanks 😊

PR #556, opened 2026-03-12
Fix RepresentationMap XML writer using wrong element for data types

### LINKED ISSUE #539: Potential bug with XML writer
Hello @javihern98 and @sosna -

I am experiencing some inconsistencies between the XML and JSON writers. Here is an example to illustrate the issue:
- I have the following RepresentationMap object created with `pysdmx`
`RepresentationMap(id='RM_Series_INDICATOR', urn='urn:sdmx:org.sdmx.infomodel.structuremapping.RepresentationMap=WB.GGH.HSP:RM_Series_INDICATOR(1.0.0)', name='Mapping Series to INDICATOR', version='1.0.0', agency='WB.GGH.HSP', source='String', target='String', maps=[ValueMap(source='per_allsp.adq_ep_preT_tot', target='SPL_TR_AMT_RD'), ValueMap(source='per_allsp.adq_ep_tot', target='SPL_TR_AMT_RD')])`

- I get different results when writing this object with `write_sdmx` if I write to XML or if I write to JSON. And I believe the XML version is incorrect. I am attaching both versions for comparison.

Image

[rep_map.json](https://github.com/user-attachments/files/25573975/rep_map.json)
[rep_map.xml](https://github.com/user-attachments/files/25573974/rep_map.xml)

javihern98:
## Summary
- Fix XML writer to use `SourceDataType`/`TargetDataType` elements when the RepresentationMap source/target is a data type (e.g. `"String"`) instead of always using `SourceCodelist`/`TargetCodelist`
- Fix XML reader to handle `SourceDataType`/`TargetDataType` elements for round-trip support
- Add test sample and assertions covering data-type RepresentationMaps

Closes #539

## Test plan
- [x] Existing `test_write_maps` round-trip test passes with the new data-type sample
- [x] New `test_write_representation_map_datatype_tags` verifies correct XML element names for both codelist and data-type cases
- [x] New reader assertions verify data-type RepresentationMap is parsed correctly
- [x] Full test suite passes (4557 tests, 100% coverage)


tests/io/test_general_reader.py
```diff
@@ -637,6 +637,27 @@ def test_read_maps():
...
     assert list(m.target) == ["MON_SAX", "Seasonally adjusted"]
 
+    # RepresentationMap 5 - RM_INDICATOR (data type source/target)
+    rep_map = result.structures[5]
+    assert isinstance(rep_map, RepresentationMap)
+    assert rep_map.id == "RM_INDICATOR"
+    assert rep_map.agency == "WB"
+    assert rep_map.version == "1.0"
+    assert rep_map.name == "Mapping Series to INDICATOR"
+    assert rep_map.source == "String"
```
sosna: Would it not be better to use DataType here, i.e. `DataType.STRING`?

sosna (commented):
It looks good, thanks, but I have a question about the possibility of using the DataType enum. We could test this by reading XML, writing JSON and then reading JSON and comparing the outputs of the two reads. What do you think?

tests/io/test_general_reader.py
javihern98: Yes, I changed the test and added a new roundtrip test as suggested above.

sosna (approved):
Looks good, thanks 😊!

PR #558, opened 2026-03-12
Accept string for dimension_at_observation parameter

### LINKED ISSUE #542: Accept string for dimension_at_observation parameter
## Description

Passing `dimension_at_observation="TIME_PERIOD"` (string instead of dict) causes an `AttributeError` because `check_dimension_at_observation()` calls `.items()` on the parameter.

## Suggested approach

Detect when a string is passed and convert it to `{short_urn: value}` for all datasets in the message. Raise `Invalid` only if the dimension specified as string is not present in any of the datasets.

## Documentation

Update `docs/howto/io_data.rst` and the `write_sdmx` docstring in `src/pysdmx/io/writer.py` to document the string shorthand.

javihern98:
## Summary
- Accept a plain string (e.g. `"TIME_PERIOD"`) for `dimension_at_observation`, expanding it to `{short_urn: value}` for all datasets in the message
- Raise `Invalid` if the dimension specified as string is not present in any dataset
- Update docstrings and documentation to describe the string shorthand

Closes #542

## Test plan
- [x] Parametrized write/read tests with string `"DIM1"` for both Generic and Structure Specific formats
- [x] Error test for invalid string dimension
- [x] All 4558 tests pass with 100% coverage
- [x] ruff and mypy clean

### DIFF
--- a/docs/howto/io_data.rst
+++ b/docs/howto/io_data.rst
@@ -175,7 +175,7 @@ A typical example to write data in Time Series with a custom header (pretty prin
         dimension_at_observation={"Dataflow=MD:TEST_DF(1.0)": "TIME_PERIOD"},
     )
 
-When all datasets share the same dimension at observation, a string shorthand can be used:
+When all datasets can use the same dimension at observation, a string shorthand can be used:
 
 .. code-block:: python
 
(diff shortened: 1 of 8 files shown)


sosna (approved):

sosna:
Nice addition, thanks 😊

PR #559, opened 2026-03-13
Add partial key rows support for SDMX-CSV writer

### LINKED ISSUE #537: Support partial key rows in SDMX-CSV writer
## Overview

When converting series-based SDMX-ML to SDMX-CSV, the writer currently flattens all series-level information into each observation row. SDMX-CSV 2.x supports "partial keys" ([section 1.3 of the field guide](https://github.com/sdmx-twg/sdmx-csv/blob/v2.1.0/data-message/docs/sdmx-csv-field-guide.md#13-mixing-rows-for-attributes-attached-to-partial-keys-with-rows-for-observations)) where series-level attributes can appear on their own rows, separate from observation rows.

## Current behavior

Given an XML input with a series containing 2 observations:
```xml
<Series FREQ="A" ... AVAILABILITY="#N/A">
    <Obs TIME_PERIOD="2003" OBS_STATUS="M" OBS_VALUE="NaN" />
    <Obs TIME_PERIOD="2004" OBS_STATUS="A" OBS_VALUE="42" />
</Series>
```

The CSV output repeats all series-level columns (including `AVAILABILITY`) for every observation row.

## Expected behavior

Support an option to write partial key rows where series-level attributes appear on a dedicated row (with empty observation-level columns), and observation data appears on separate rows with empty series-level attribute columns.

## Context

Raised during review of PR #499 (issue #396). Rated as "normal" priority.

javihern98:
## Summary
- Adds `partial_keys` parameter to SDMX-CSV 2.0 and 2.1 writers, enabling partial key rows per SDMX-CSV 2.x section 1.3
- Series-level and group-level attributes appear on their own rows with only attachment dimensions filled; observation rows omit those attributes
- Raises `Invalid` when `partial_keys=True` is used without a `Schema` structure

## Test plan
- [x] Partial key rows with multiple observations in the same series (both sdmx20 and sdmx21 writers)
- [x] All obs-level attributes: `partial_keys=True` is a no-op
- [x] Empty attribute values produce no partial key rows
- [x] `partial_keys=True` without Schema raises `Invalid`

Closes #537


tests/io/csv/sdmx20/writer/samples/reference_partial_keys.csv
```diff
@@ -0,0 +1,4 @@
+STRUCTURE,STRUCTURE_ID,ACTION,DIM1,DIM2,ATT1,ATT2,OBS_VALUE,TIME_PERIOD
```
sosna: This would be slightly easier to understand if all dimensions (i.e. DIM1, DIM2, TIME_PERIOD) were next to each other. My first reaction when seeing this was that the example was wrong, until I realized that TIME_PERIOD (i.e. the OBS dimension) had been pushed as the last column.

tests/io/csv/sdmx21/writer/samples/reference_partial_keys.csv
```diff
@@ -0,0 +1,4 @@
+STRUCTURE,STRUCTURE_ID,ACTION,DIM1,DIM2,ATT1,ATT2,OBS_VALUE,TIME_PERIOD
```
sosna: This would be slightly easier to understand if all dimensions (i.e. DIM1, DIM2, TIME_PERIOD) were next to each other. My first reaction when seeing this was that the example was wrong, until I realized that TIME_PERIOD (i.e. the OBS dimension) had been pushed as the last column.

sosna (changes requested):
It's really nice and we are almost there 😊. However, there is one issue remaining.

I sent an email on the 19th of February this year with sample files.

I then also have this script:

```python
from pysdmx.io import get_datasets, write_sdmx

data = get_datasets("data.in.xml", "structure.json", validate=False)
write_sdmx(data, Format.DATA_SDMX_CSV_2_1_0, "data.out.csv", partial_keys=True)
```

Using the sample files I sent last month, the CSV output is as follows:

```csv
STRUCTURE,STRUCTURE_ID,ACTION,FREQ,DER_TYPE,DER_INSTR,DER_RISK,DER_REP_CTY,DER_SECTOR_CPY,DER_CPC,DER_SECTOR_UDL,DER_CURR_LEG1,DER_CURR_LEG2,DER_ISSUE_MAT,DER_RATING,DER_EX_METHOD,DER_BASIS,AVAILABILITY,TIME_PERIOD,OBS_STATUS,OBS_VALUE,OBS_CONF,DECIMALS,UNIT_MEASURE,UNIT_MULT
dataflow,BIS:BIS_DER(1.0),M,A,U,A,B,5J,A,1E,A,HKD,TO1,A,A,3,C,#N/A,,,,,,,
dataflow,BIS:BIS_DER(1.0),M,A,U,A,B,5J,A,1E,A,HKD,TO1,A,A,3,C,,2003,M,NaN,,None,None,None
dataflow,BIS:BIS_DER(1.0),M,A,U,A,B,5J,A,1E,A,HKD,TO1,A,A,3,C,,2004,A,42,#N/A,None,None,None
```

As can be seen:
- We now have a nice series row (the first row), i.e. the equivalent of the XML `Series` element, which is great 😊.
- However, there is an issue with the values of the last 3 attributes in the last 2 rows, i.e. the rows that represent observations. The value is `None` but, instead, I guess we want the cells to be empty.

tests/io/csv/sdmx20/writer/samples/reference_partial_keys.csv
javihern98: Reordered columns when partial_keys=True (for the other cases requires the structure to be present, which is handled in #349 )

tests/io/csv/sdmx21/writer/samples/reference_partial_keys.csv
javihern98: Same as above

javihern98:
> However, there is an issue with the values of the last 3 attributes in the last 2 rows, i.e. the rows that represent observations. The value is None but, instead, I guess we want the cells to be empty.

@sosna I fixed this as well, thanks for spotting this :blush:

sosna (approved):
Looks great, thanks 😊!

tests/io/csv/sdmx20/writer/samples/reference_partial_keys.csv
sosna: Thanks 😊

Issue #560, opened 2026-03-13
Migrate from NumPy to PyArrow dtypes for the data extra


javihern98:
# Migrate from NumPy to PyArrow dtypes for the `data` extra

Closes #509

## Summary

Replace numpy with pyarrow as the dtype backend for the `data` extra. Make `PandasDataset` automatically apply pyarrow-backed dtypes based on the SDMX Schema when present in `dataset.structure`. When no Schema is present (structure is a string URN), all DataFrame columns should be `string[pyarrow]`.

This also resolves #509 (pyarrow dictionary dtype causing `ArrowNotImplementedError` in the XML structure-specific writer). By standardizing all DataFrame columns to well-known pyarrow types via `PandasDataset`, dictionary-encoded columns with null values will no longer reach the writer — `PandasDataset` will always produce `pa.string()` columns (not dictionary types), so `.astype(str)` in the writer works correctly.

## Motivation

- numpy is barely used in the codebase (one `np.nan` reference in the XML structure-specific reader)
- DataFrames from readers are currently all-string with no schema-driven typing
- The existing `to_pandas_type`/`to_pandas_schema` utilities in `toolkit/pd` map SDMX DataTypes to numpy types but are unused by any reader
- pyarrow provides a modern, memory-efficient dtype backend with native nullable types (no need for the required/nullable distinction numpy requires)

### New PyArrow Type Mapping Functions

Add new functions in `toolkit/pd/__init__.py` alongside the existing numpy-based ones (which remain unchanged):

```python
def to_pyarrow_type(comp: Component) -> pd.ArrowDtype
def to_pyarrow_schema(components: Iterable[Component]) -> Dict[str, pd.ArrowDtype]
```

Functions return `pd.ArrowDtype` objects (not strings), because `astype('string[pyarrow]')` produces a pandas `StringDtype` instead of `pd.ArrowDtype`. Using `pd.ArrowDtype(pa.string())` ensures consistent ArrowDtype across all columns.

**Complete type mapping for all SDMX DataTypes:**

| SDMX DataType | PyArrow type | Resulting dtype |
| --- | --- | --- |
| SHORT | `pa.int16()` | `int16[pyarrow]` |
| INTEGER | `pa.int32()` | `int32[pyarrow]` |
| LONG | `pa.int64()` | `int64[pyarrow]` |
| COUNT | `pa.int64()` | `int64[pyarrow]` |
| BIG_INTEGER | `pa.string()` | `string[pyarrow]` |
[...]

PR #562, opened 2026-03-19
Add a SQL Server toolkit to pysdmx

### LINKED ISSUE #315: Add mappers to SQL
pysdmx offers classes representing data filters or queries (`pysdmx.api.dc.query`). It would be nice to have utility functions to easily map these to SQL

sosna:
This PR adds a SQL Server toolkit to pysdmx.

With the toolkit, you can:

- Create tables in SQL Server, based on structural metadata, using the `create_table` function. While only the structural metadata is required for the function to work, additional parameters can be used to tweak the creation process (like defining the columns to be indexed, the ones to be used as primary key, etc.).
- Convert pysdmx.api.dc.query filters into a SQL SELECT prepared statement, using the `get_select_statement` function.
- Additional functions are available but these are more lower-level.

Close #315


src/pysdmx/toolkit/pd/__init__.py
```diff
@@ -60,6 +58,11 @@ def to_pandas_type(comp: Component) -> str:
     """
     if comp.enumeration:
         return "category"
+    elif comp.dtype == DataType.INCREMENTAL:
+        if comp.facets and isinstance(comp.facets.interval, float):
+            return "float32" if comp.required else "Float32"
+        else:
+            return "int32" if comp.required else "Int32"
```
javihern98, line 65: Why is this if needed? Is it to handle the nullable data types in Pandas? We aim to change them to pyarrow types in #560, maybe its best to handle the mapping always to pyarrow data types already

javihern98 (approved):
Looks good, thanks! 😊  Just a small question over the nullable data types in the Pandas mapping included in this PR

src/pysdmx/toolkit/pd/__init__.py
sosna: Thanks, Javier!

I added this because of the new incremental data type (needed because of the IDENTITY type in SQL Server). Increments in SDMX can either be an integer or a float, and the component itself can be required (or not). 

As I added this new DataType, I thought I also needed to update this mapping between SDMX data types and Pandas data types, else the type in Pandas would have been to "string".

 Now, re. Pandas vs. pyarrow types, I guess we'll want to migrate all types at once, and not some now, and some later?

Does that make sense?

javihern98: Hi Xavier, makes sense, let's proceed with the merge then😊

sosna: Perfect! Thanks, @javihern98 😊!

PR #570, opened 2026-04-24
Provide connectors for the upcoming SDMX Data Discovery and Retrieval profile

### LINKED ISSUE #176: Implement BIS Data Connector
The purpose of this task is to use the SDMX-REST query builders (in the `pysdmx.api.qb` package), to build a generic SDMX Connector (based on the API defined in `pysdmx.api.dc.Connector`) and use it to discover and retrieve data from the BIS web service.

sosna:
This PR provides two connectors for the upcoming SDMX Data Discovery and Retrieval profile.

Please note that the connectors are experimental only, as the profile is currently being defined in the SDMX-TWG and still subject to change.

Close #176


javihern98:
### Code review

Found 3 issues:

1. `_sanitize_query` URL-encodes characters that are part of valid SDMX-REST query syntax (`[`, `]`, `:`, `+`, `*`, `,`), and is applied to the entire query (path + query string) inside both `__fetch` and `__stream` for sync and async services. Existing component-filter URLs like `c[FREQ]=ne:M`, `c[TIME_PERIOD]=ge:2011+le:2022`, wildcard data paths like `/data/*/*/*/*/*`, and IN-list values like `A,M` are all silently corrupted. The new streaming tests only exercise a bare `DataQuery(DataContext.DATAFLOW, "BIS", "CBS")` with no component filters, so this regression is not caught. This affects every query type going through `__fetch` (`data`, `structure`, `schema`, `availability`, `reference_metadata`).

https://github.com/bis-med-it/pysdmx/blob/0cc4ea8f4702c0a744eae09755741bdf9ff43d81/src/pysdmx/api/qb/service.py#L419-L429

2. `parse_flow_urn` hardcodes `sdmx_type=\"Dataflow\"` and is wired in as the final fallback inside the generic `parse_urn`. The regex `^(.*):(.*)\((.*)\)$` matches any `agency:id(version)` string, so any non-dataflow short URN reaching this branch (e.g. shorthand for a codelist or concept scheme) is silently mistyped as a Dataflow rather than raising `Invalid`. Consider scoping `parse_flow_urn` to dataflow-specific call sites instead of `parse_urn`'s generic fallback chain.

https://github.com/bis-med-it/pysdmx/blob/0cc4ea8f4702c0a744eae09755741bdf9ff43d81/src/pysdmx/util/__init__.py#L86-L99

3. `PandasConnector` docstring states the targeted service must return data in SDMX-CSV v2.0.0, but the constructor configures `RestService` with `DataFormat.SDMX_CSV_1_0_0`. The sibling `SdmxConnector` in `_rest.py` correctly uses `SDMX_CSV_2_0_0`, suggesting the version mismatch here is unintentional rather than a deliberate fallback.

https://github.com/bis-med-it/pysdmx/blob/0cc4ea8f4702c0a744eae09755741bdf9ff43d81/src/pysdmx/api/dc/pd.py#L26-L46

sosna:
Thanks, @javihern98. For the first 2, I will have a look today 😊. For the 3rd one, please have a look at the email exchange with Antonio on Friday (i.e. this is a temporary thing).

javihern98:
> Thanks, @javihern98 . For the first 2, I will have a look today 😊. For the 3rd one, please have a look at the email exchange with Antonio on Friday (i.e. this is a temporary thing).

Hi @sosna, noted, sorry for the automatic posting on the PR by Claude, it was meant to be a local review process, but seems right in any case :blush:

sosna:
I agree, it seems to be good stuff indeed, thanks 😊

sosna:
The second comment above (about parse_flow_urn) has been addressed in commits https://github.com/bis-med-it/pysdmx/pull/570/commits/24a6a88a381db24c33f6b8fb0630a7d54f9a78d8 and https://github.com/bis-med-it/pysdmx/pull/570/commits/7241b2de89a16ce23f3728342b5fcc8d25ac2341.

javihern98 (approved):
Looks good, thanks! :blush:

PR #574, opened 2026-04-29
Move per-extra type stubs to extras-dev group

javihern98:
## Summary

Decouples per-extra type stubs from the default `dev` install and replaces the hand-maintained `WANT` dict in the extras workflow with a dynamic read of `[project.optional-dependencies]`.

- New non-optional `extras-dev` Poetry group holds the per-extra type stubs (`lxml-stubs`, `types-xmltodict`, `types-python-dateutil`, `pyarrow-stubs`, `pandas-stubs`, `types-jsonschema`); the `dev` group keeps only generic tooling. This stops `pyarrow-stubs` / `pandas-stubs` from transitively pulling pyarrow/pandas into every install.
- `pyarrow` constraint in the `data` and `all` extras and in `extras-dev` now splits by `python_version` (matching the existing `numpy` and `pandas-stubs` pattern); combined with `vtlengine`'s `<20` bound, the lock collapses to a single pyarrow entry.
- `extra.yml` install step adds `--without extras-dev` so per-extra runs don't pull stubs.
- `extra.yml` verify step replaces the hardcoded `WANT` dict with a dynamic read of `[project.optional-dependencies]` from `pyproject.toml`, walking transitive deps via `importlib.metadata.requires` so legitimate transitives (e.g. `vtlengine` pulling `pandas`) aren't flagged. Adding a new extra to `pyproject.toml` no longer requires updating the workflow.
- Adds `.venv/` to `.gitignore`.


pyproject.toml
```diff
@@ -35,12 +35,16 @@ documentation = "https://bis-med-it.github.io/pysdmx"
 "Bug Tracker" = "https://bis-med-it.github.io/pysdmx/issues"
 
 [project.optional-dependencies]
-data = ["pandas>=2.1.4,<3.0", "pyarrow>=14.0,<20.0", "numpy>=2.0.2,<2.1; python_version < '3.10'", "numpy>=2.1.0,<2.3; python_version >= '3.10'"]
+data = ["pandas>=2.1.4,<3.0",
+    "pyarrow>=14.0,<20.0; python_version < '3.10'", "pyarrow>=14.0,<25.0; python_version >= '3.10'",
```
sosna, line 39: Do we need an upper bound (i.e., `<25.0`) for the most recent versions or is this unnecessarily limiting users of our libraries?

pyproject.toml
```diff
@@ -35,12 +35,16 @@ documentation = "https://bis-med-it.github.io/pysdmx"
...
+    "numpy>=2.0.2,<2.1; python_version < '3.10'", "numpy>=2.1.0,<2.3; python_version >= '3.10'"]
```
sosna, line 40: Same question as above for the upper bound on numpy (<2.3)...

sosna (approved):
It looks good but I have a question on the upper bound. Could you please clarify? Thanks!

pyproject.toml
javihern98: It is to prevent unexpected behaviours on latest versions of numpy as pandas is migrating all dtypes to pyarrow and slowly increase it. As it is just a Pandas dependency, I am really concerned about issues on installation.

Furthermore, numpy binaries could behave wrong if a gcc version is not supported at some point, they really have a bad habit of breaking installations even on patch versions because they increase the gcc or python version requirements

pyproject.toml
javihern98: It is to prevent unexpected behaviours on latest versions of pyarrow as pandas is migrating all dtypes to pyarrow and slowly increase it. As it is just a Pandas dependency, I am really concerned about issues on installation with major versions.

pyproject.toml
javihern98: Numpy versions are supported for the three last minor versions

https://endoflife.date/numpy

PR #576, opened 2026-04-30
Set default version in the FMR API from `+` to `~` as per the SDMX-REST spec

### LINKED ISSUE #575: REST-API calls without version use + wildcard, fails for 2-part versions in FMR 12
## Summary

When no version is specified, pysdmx appends `/+/` to SDMX REST v2 URLs. Per the [SDMX REST API spec](https://github.com/sdmx-twg/sdmx-rest/blob/master/doc/querying_versions.md), `+` means "latest stable semantic version" and **only matches 3-part versions** (`x.y.z`) in the latest builds of FMR 12. Structures with 2-part versions (e.g. `1.0`) are excluded, resulting in a 404.

The version segment should be omitted (defaulting to `~`, which matches all version formats) or `~` should be used explicitly.

## Reproduction

Observed against FMR 12.0.0-beta.24, which implements the above SDMX REST API spec for `+`.

```python
from pysdmx import create_client

client = create_client("https://example.com/sdmx/v2", "SDMX_REST_2")

# Fails (404) on FMR 12 — generates URL: .../codelist/DEMO/CL_REF_AREA/+/
codes = client.get_codes("DEMO", "CL_REF_AREA")

# Works — generates URL: .../codelist/DEMO/CL_REF_AREA/1.0/
codes = client.get_codes("DEMO", "CL_REF_AREA", "1.0")
```

## Affected methods

Any method that constructs a structure query without an explicit version, including at least:
- `get_codes()`
- `get_code_map()`

## Suggested fix

When the user does not supply a version, omit the version path segment entirely (the SDMX REST v2 API defaults to `~`) instead of sending `+`.

## Environment

- Tested against FMR 12.0.0-beta.24 (spec-compliant: `+` rejects 2-part versions)

sosna:
Close #575


javihern98 (approved):
Looks good, thanks! :blush:

PR #502, opened 2026-01-12
Fix IndexError in generic writer when series attributes vary

ruizmaa:
This PR fixes #500, a bug where the generic writer raised an `IndexError: list index out of range` when attributes were not constant within a series.

Eg:
A dataset with Dimension Country and Series Attribute Currency

| Country (Dim) | Currency (Attr) | Value |
| :--- | :--- | :--- |
| ES | EUR | 100 |
| ES | USD | 110 |

The writer initialized the list with only dimensions (`Country=ES`), creating a list of length 1. While, the processing loop grouped data by dimensions + attributes, resulting in 2 groups (`ES-EUR` and `ES-USD`). So there was a missmatch

## Changes
Updated `__series_processing` in `generic.py` to include `series_att_codes` in both the `sort_values` and the `data_dict` initialization. So if attributes vary, the writer splits the data into multiple `<gen:Series>` blocks instead of failing.

## Test
Added test `test_generic_writer_varying_attributes` to ensure correct XML generation when this situation happens

Issue #500, opened 2026-01-07
IndexError in Generic format writer with `dimension_at_observation`


ruizmaa:
When writing SDMX-ML 2.1 Generic format data with `dimension_at_observation`, the writer crashes with an `IndexError: list index out of range` if series-level attributes have different values across observations within the same series.

```
IndexError: list index out of range
```

At `src/pysdmx/io/xml/sdmx21/writer/generic.py`, line 350:
```python
data_dict["Series"][0]["Obs"] = obs_data.to_dict(orient="records")
```

## Reproduce bug:

```python
from pathlib import Path
from pysdmx.io import read_sdmx
from pysdmx.io.format import Format
from pysdmx.io.pd import PandasDataset
from pysdmx.io.reader import get_datasets
from pysdmx.io.writer import write_sdmx

# Load schema
struct_path = Path("tests/io/samples/datastructure.xml")
msg = read_sdmx(struct_path)
dsds = msg.get_data_structure_definitions()
schema = dsds[0].to_schema()

# Load CSV data with varying series attributes
csv_data= """STRUCTURE,STRUCTURE_ID,ACTION,FREQ,DER_TYPE,DER_INSTR,DER_RISK,DER_REP_CTY,DER_SECTOR_CPY,DER_CPC,DER_SECTOR_UDL,DER_CURR_LEG1,DER_CURR_LEG2,DER_ISSUE_MAT,DER_RATING,DER_EX_METHOD,DER_BASIS,AVAILABILITY,COLLECTION,TIME_FORMAT,TIME_PERIOD,OBS_STATUS,OBS_VALUE,OBS_CONF
dataflow,BIS:BIS_DER(1.0),M,A,U,A,B,5J,A,1E,A,HKD,TO1,A,A,3,C,#N/A,A,,,,,
dataflow,BIS:BIS_DER(1.0),M,A,U,A,B,5J,A,1E,A,HKD,TO1,A,A,3,C,,,,2003,M,NaN,
dataflow,BIS:BIS_DER(1.0),M,A,U,A,B,5J,A,1E,A,HKD,TO1,A,A,3,C,,,,2004,,14206.490766,#N/A
dataflow,BIS:BIS_DER(1.0),M,A,U,A,B,5J,A,1E,A,HKD,TO1,A,A,3,C,,,P1M,,,,"""

datasets = get_datasets(csv_data)
dataset = PandasDataset(data=datasets[0].data, structure=schema)

# This crashes with IndexError
[...]

PR #499, opened 2026-01-07
Enforce Schema validation and preserve null values

ruizmaa:
This PR addresses issue #396 from scratch, superseding the PR #470 

### 1. Schema enforcement

- **Schema requirement**: Writing operations now strictly require a `Schema` object in `dataset.structure`. `validate_schema_exists` have been implemented to ensure it, raising an `Invalid` error if there is no schema in the dataset to convert.

### 2. No longer null conversion to empty strings

- Removed `.replace({"nan": "", "<NA>": ""})`, `.fillna("")` or `.replace("nan", "")` from the writers.
- Replaced `pd.read_csv(na_values=[""], ...)` to  `pd.read_csv(na_values=[], ...)` from the readers.

### 3. Normalize nulls when reading

- Keeped XML readers (`__ss_aux_reader.py`) null normalization `.replace(np.nan, "")`
- Updated CSV readers null replacement from `.replace({"nan": "", "<NA>": ""})` to `.replace({"nan": "NaN", "<NA>": "NaN"})`
- Added null normalization in generic reader as in non generic: `.replace(np.nan, "")` in `__reading_generic_groups()`, `__reading_generic_series()` and `__reading_generic_all()`

### 4. Deleted `__remove_optional_attributes_empty_data()` and added `transform_dataframe_for_writing()`

Now `transform_dataframe_for_writing()` recives a `pd.DataFrame` and a `Schema` and converts nulls and empty values based on if it's required or not.

It relies on auxiliary functions:

- `_get_value_to_write`: decides based the value's `dtype` and if it's `required` whether if it should be written (if it has any value or it's null), skip it (if it's not required and has an empty value `""`) or write null (if its originally null (`NaN` or `#N/A`) or if it's required and has an empty value `""`). It uses more auxiliary functions
  - `_is_null_value`
  - `_is_nullable_integer_dtype`: a helper function to correctly write numbers format (avoid `1` to be written as `1.0`, `1.0` as 1`)
  - `_get_null_representation`: returns `"NaN"` for numeric and `"#N/A"` for the others based on the `NUMERIC_TYPES` set:

```python
NUMERIC_TYPES = {
    DataType.BIG_INTEGER,
    DataType.COUNT,
    DataType.DECIMAL,
    DataType.DOUBLE,
    DataType.FLOAT,
    DataType.INTEGER,
    DataType.LONG,
    DataType.SHORT,
}
```

So the new behavior is:

| Original file value                        | Read value | Written value (CSV)                                   | Written value (XML)                                   |
[...]

PR #590, opened 2026-05-11
Fix duplicate <Series>/<Group> for series-attached attributes

### LINKED ISSUE #588: Structure-specific SDMX-ML writer emits duplicate <Series> when an optional series attribute is "#N/A"
## Summary

Discovered while testing #499 (see [review comment](https://github.com/bis-med-it/pysdmx/pull/499#issuecomment-4344758958)).
When writing a dataset to `Format.DATA_SDMX_ML_2_1_STR` (Structure-Specific SDMX-ML 2.1), if a series
contains both observations and an optional series-level attribute (e.g. `AVAILABILITY`) holding an
explicitly-missing value (`#N/A`), the writer emits the same series key twice:

- one `<Series>` element carrying the observations (with the series attribute omitted), and
- one `<Series>` element carrying the series attribute (self-closing, no observations).

The round trip is SDMX-CSV → SDMX-ML. The reverse direction (SDMX-ML → SDMX-CSV) works correctly.

This is **not** a regression of #492 / PR #518 — that fixed the self-closing rendering for series
that have *no* observations. The current bug is that a series that *does* have observations is being
split apart by the writer's groupby/dedup over `series_codes`.

## Reproduction

Round-trip the SDMX-CSV input used in #492 back to `Format.DATA_SDMX_ML_2_1_STR`. The output contains
two `<Series>` elements for the same dimension key:

```xml
<Series FREQ="A" DER_TYPE="U" DER_INSTR="A" DER_RISK="B" DER_REP_CTY="5J"
        DER_SECTOR_CPY="A" DER_CPC="1E" DER_SECTOR_UDL="A" DER_CURR_LEG1="HKD"
        DER_CURR_LEG2="TO1" DER_ISSUE_MAT="A" DER_RATING="A" DER_EX_METHOD="3"
        DER_BASIS="C">
    <Obs TIME_PERIOD="2003" OBS_VALUE="NaN" OBS_STATUS="M" />
    <Obs TIME_PERIOD="2004" OBS_VALUE="42" OBS_STATUS="A" OBS_CONF="#N/A" />
</Series>
<Series FREQ="A" DER_TYPE="U" DER_INSTR="A" DER_RISK="B" DER_REP_CTY="5J"
        DER_SECTOR_CPY="A" DER_CPC="1E" DER_SECTOR_UDL="A" DER_CURR_LEG1="HKD"
        DER_CURR_LEG2="TO1" DER_ISSUE_MAT="A" DER_RATING="A" DER_EX_METHOD="3"
        DER_BASIS="C" AVAILABILITY="#N/A" />
```

## Expected

A single `<Series>` element per dimension key, carrying `AVAILABILITY="#N/A"` and both `<Obs>` children.

## Analysis (suggested starting point)

Likely in `src/pysdmx/io/xml/__write_structure_specific_aux.py`, function `__series_processing`
(lines 243-313). Both the dedup at lines 305-308 and the
`data.groupby(by=series_codes, dropna=False)` call at line 252 partition the input by `series_codes`.
If `AVAILABILITY` is present in `series_codes`, the rows where `AVAILABILITY="#N/A"` (the
series-attribute row in the source CSV) and the rows where `AVAILABILITY=""` (the observation rows)
end up in separate groups, so the writer emits two `<Series>` elements.

Worth verifying whether reading the SDMX-CSV should forward-fill the series attribute onto the
observation rows before writing, or whether `__series_processing` should compute the series-level
attribute value per dimension group rather than including it in the groupby key.

A regression test should cover: one series, one optional series-level attribute, value `#N/A`,
plus at least one observation, format `DATA_SDMX_ML_2_1_STR`.

javihern98:
Closes #588.

Discovered by @sosna while reviewing #499 — the structure-specific and
generic SDMX-ML writers were grouping rows by `series_codes` (dimensions
*and* series-attached attributes) and dedup'ing on the same key. A series
where one row carried `AVAILABILITY="#N/A"` and the other rows left it
empty was therefore split into two `<Series>` elements (one with obs and
no attribute, one self-closing with the attribute). The structure-specific
`<Group>` emission had the same defect for group-attached attributes.

## Changes

- **`src/pysdmx/io/xml/__write_structure_specific_aux.py`** — group/dedup
  `__series_processing` by dimensions only; derive each series-attached
  attribute via `_first_non_empty` over the group rows. Filter the obs
  list to rows whose observation dimension is non-empty so that
  series-attribute-only rows are not emitted as `<Obs />`. Same fix
  applied to `__group_processing` for group-attached attributes.
  `__has_valid_obs` is removed (now unreachable) and `__format_ser_str`
  is promoted to module level to keep the complexity below the project
  cap.
- **`src/pysdmx/io/xml/sdmx21/writer/generic.py`** — same fix in
  `__series_processing`: group by dims only, derive each series attr
  per group. The existing obs-row filter in `__format_ser_str` already
  handles the series-attribute-only rows.
- Both fixes carry over to SDMX-ML 3.0 and 3.1 (shared aux module) and
  to the Time Series wrappers (which delegate to the structure-specific
  and generic writers).

## Tests

Added four regression tests that fail on `develop` HEAD and pass with
the fix:

- `test_str_spec_series_attr_only_row_merges_with_obs` — 2.1 Structure-Specific
- `test_generic_series_attr_only_row_merges_with_obs` — 2.1 Generic
- `test_str_spec_group_attr_mixed_empty_na_emits_single_group` — 2.1 Group
- `test_str_spec_series_attr_only_row_merges_with_obs_30` — 3.0 Structure-Specific

`test_generic_writer_varying_attributes` (added in #502 for #500) was
updated to reflect the SDMX-correct semantics: a series-attached
attribute is constant across a series, so the writer now emits one
`<gen:Series>` per series key with the first non-empty value. The
crash regression guard from #500 remains in place.

Existing regression guards in the writer test suites
(`test_series_format_with_series_no_obs`,
`test_write_generic_no_groups_series_attributes`,
`test_write_structure_specific_optional_attributes`,
`test_data_write_nullable_nulltypes`, …) all still pass.


tests/io/xml/sdmx21/writer/test_data_writing.py
```diff
@@ -578,11 +585,9 @@ def test_generic_writer_varying_attributes(header, content):
...
 
     assert result is not None
-    assert "gen:Series" in result
-    assert "gen:Obs" in result
-    assert result.count("<gen:Series>") == 2
-    assert "A" in result
-    assert "B" in result
+    assert result.count("<gen:Series>") == 1
+    assert result.count("<gen:Obs>") == 2
+    assert '<gen:Value id="ATT1" value="A"/>' in result
```
sosna, line 618: Are we sure this is correct behavior? Why assuming the first value is correct? As PEP 20 says: "In the face of ambiguity, refuse the temptation to guess." 😁

Maybe we'd be better off raising an exception?

tests/io/xml/sdmx30/writer/test_data_writing.py
```diff
@@ -492,3 +494,60 @@ def test_explicit_null_sentinels_preserved_in_xml_output():
     assert 'OBS_VALUE="NaN"' in xml_out
     assert 'OBS_VALUE="42"' in xml_out
     assert 'O_ATT="#N/A"' in xml_out
+
+
+def test_str_spec_series_attr_only_row_merges_with_obs_30():
+    """Issue #588: SDMX-ML 3.0 series-attr-only row must merge with obs."""
```
sosna: Is the doc string correct? Based on the asserts below, I would assume that the series-level attribute is attached to a Series element rather than being merged with the obs?

tests/io/xml/sdmx21/writer/test_data_writing.py
javihern98: Yes, I changed the code to raise an Invalid error if that happens. PEP 20 is always a good call :blush:

tests/io/xml/sdmx30/writer/test_data_writing.py
javihern98: I changed the docstring, the fix collapses the duplicate <Series> into one and attaches the attribute on the <Series> element itself, not on <Obs>. Happy to change it further if needed

sosna (approved):
Looks good, thanks 😊

PR #557, opened 2026-03-12
Add writer and reader support for Time Series formats (STRTS / GENTS)

javihern98:
## Summary

Closes #540

- Register `DATA_SDMX_ML_2_1_GENTS` and `DATA_SDMX_ML_2_1_STRTS` in both the writer and reader pipelines
- Reuse existing generic and structure-specific writers/readers — TS wrappers default `dimension_at_observation` to `TIME_PERIOD` and use the correct SDMX-ML 2.1 Time Series message types (`GenericTimeSeriesData` / `StructureSpecificTimeSeriesData`)
- Add format auto-detection for TS messages in the reader
- Update documentation (`general_writer.rst`, `sdmx_ml.rst`, `io_data.rst`)

## Test plan

- [x] 12 new tests covering root element names, TIME_PERIOD defaults, explicit overrides, round-trip write/read, and `write_sdmx()` integration
- [x] Full test suite passes (4567 tests, 100% coverage)
- [x] ruff format, ruff check, mypy all clean

PR #591, opened 2026-05-11
Accept string shortcut for dimension_at_observation in TS writers

### LINKED ISSUE #589: dimension_at_observation string shortcut crashes Time Series writers (STRTS, GENTS)
## Summary

Discovered while testing #499 (see [review comment](https://github.com/bis-med-it/pysdmx/pull/499#issuecomment-4344758958)).
The `dimension_at_observation` string shortcut introduced in #542 / #558 is not honoured by the
two Time Series writers added in #540 / #557. Calling either with a plain string crashes with
`AttributeError: 'str' object has no attribute 'items'`.

Affected formats:
- `Format.DATA_SDMX_ML_2_1_STRTS` (`src/pysdmx/io/xml/sdmx21/writer/structure_specific_ts.py`)
- `Format.DATA_SDMX_ML_2_1_GENTS` (`src/pysdmx/io/xml/sdmx21/writer/generic_ts.py`)

The non-TS writers (`generic.py`, `structure_specific.py`) handle the shortcut correctly.

## Reproduction

```python

# Crashes:
write_sdmx(
    datasets,
    format=Format.DATA_SDMX_ML_2_1_STRTS,
    dimension_at_observation="TIME_PERIOD",
)
```

Traceback:

```
File "src/pysdmx/io/writer.py", line 192, in write_sdmx
    return writer(**args)
File "src/pysdmx/io/xml/sdmx21/writer/structure_specific_ts.py", line 48, in write
    for k, v in dimension_at_observation.items()
AttributeError: 'str' object has no attribute 'items'
```

The same call against `Format.DATA_SDMX_ML_2_1_GENTS` fails identically at
`src/pysdmx/io/xml/sdmx21/writer/generic_ts.py:46`.

Workaround — the explicit dict form works for both:

```python
dimension_at_observation={"Dataflow=BIS:BIS_DER(1.0)": "TIME_PERIOD"}
```

## Analysis

Both TS wrappers declare `dimension_at_observation: Optional[Dict[str, str]]` and call `.items()`
directly before delegating to the base writer. They do not invoke the shared
`check_dimension_at_observation()` helper in `src/pysdmx/io/xml/__write_data_aux.py` that the non-TS
writers use to normalise a string shortcut to the per-dataset dict.

For comparison:
- `generic.py`: type hint `Optional[Union[str, Dict[str, str]]]`, calls `check_dimension_at_observation()`.
- `structure_specific.py`: same pattern.

## Suggested fix

Align both TS wrappers with the non-TS writers:
1. Widen the type hint to `Optional[Union[str, Dict[str, str]]]`.
2. At the top of `write()`, when `dimension_at_observation` is a string, validate it equals
   `"TIME_PERIOD"` and expand it to `{ds.short_urn: "TIME_PERIOD" for ds in datasets}` before
   running the existing per-dataset validation loop.
3. Add regression tests calling each TS writer with the string shortcut.

javihern98:
Closes #589.

The SDMX-ML 2.1 Time Series writers (STRTS, GENTS) inherited a stricter
`dimension_at_observation: Optional[Dict[str, str]]` signature from when
they were added in #557 (cr-540), and were not updated when #558 (cr-542)
introduced the string shortcut for the non-TS writers. Passing a plain
string crashed both with `AttributeError: 'str' object has no attribute
'items'`.

## Changes

- **`src/pysdmx/io/xml/sdmx21/writer/structure_specific_ts.py`** and
  **`src/pysdmx/io/xml/sdmx21/writer/generic_ts.py`** — widen the type
  hint to `Optional[Union[str, Dict[str, str]]]` and add an
  `elif isinstance(..., str)` branch that expands the string to a
  per-dataset dict (the same shape the `None` branch already produces).
  The existing "must be TIME_PERIOD" validation then runs unchanged, so
  passing `"FREQ"` still raises `Invalid` with the same message.

The fix is intentionally minimal: it keeps the TS wrapper's
responsibilities small and matches its existing structure rather than
delegating to `check_dimension_at_observation()` (which expects a
`Dict[str, PandasDataset]` and adds a "dimension must exist" check the
TS wrapper does not currently perform).

## Tests

Added four regression tests to
`tests/io/xml/sdmx21/writer/test_data_writing_ts.py`, all of which
fail on `develop` HEAD with the original `AttributeError`:

- `test_generic_ts_accepts_time_period_string` — happy path for GENTS
- `test_structure_specific_ts_accepts_time_period_string` — happy path for STRTS
- `test_generic_ts_rejects_non_time_period_string` — non-TIME_PERIOD string raises `Invalid`
- `test_structure_specific_ts_rejects_non_time_period_string` — same for STRTS

Existing regression guards keep passing: `test_generic_ts_defaults_time_period`,
`test_generic_ts_explicit_time_period`, `test_generic_ts_rejects_non_time_period`,
their STRTS counterparts, the round-trip tests, and the `write_sdmx` parametrized
format test.

### DIFF
--- a/src/pysdmx/io/xml/sdmx21/writer/generic_ts.py
+++ b/src/pysdmx/io/xml/sdmx21/writer/generic_ts.py
@@ -16,7 +16,7 @@ def write(
     output_path: Optional[Union[str, Path]] = None,
     prettyprint: bool = True,
     header: Optional[Header] = None,
-    dimension_at_observation: Optional[Dict[str, str]] = None,
+    dimension_at_observation: Optional[Union[str, Dict[str, str]]] = None,
 ) -> Optional[str]:
     """Write data to SDMX-ML 2.1 Generic Time Series format.
 
@@ -30,8 +30,9 @@ def write(
         prettyprint: Prettyprint or not.
         header: The header to be used (generated if None).
         dimension_at_observation:
-          The mapping between the dataset and the dimension at observation.
-          Defaults to TIME_PERIOD for all datasets.
+          The dimension at observation, either as a string applied to all
+          datasets or a dict mapping short URNs to dimension IDs. Defaults
+          to TIME_PERIOD for all datasets.
 
     Returns:
         The XML string if path is empty, None otherwise.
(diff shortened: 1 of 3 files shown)


sosna (approved):
Looks good, thanks 😊

PR #593, opened 2026-05-12
Support stable semver version strings in the FMR clients

### LINKED ISSUE #592: Special + character for version is poorly supported
Using + for the version when retrieving structures is poorly supported, as can be seen with the stack trace below:

```
Traceback (most recent call last):   File "<string>", line 1, in <module>   File "C:\Users\fl007141\envs\medconn\Lib\site-packages\pysdmx\api\fmr\__init__.py", line 652, in get_dataflow_details     return super()._out(            ^^^^^^^^^^^^^   File "C:\Users\fl007141\envs\medconn\Lib\site-packages\pysdmx\api\fmr\__init__.py", line 102, in _out     return decode(response, type=typ).to_model(*params)            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^   File "C:\Users\fl007141\envs\medconn\Lib\site-packages\pysdmx\io\json\fusion\messages\dataflow.py", line 77, in to_model     df = list(          ^^^^^ IndexError: list index out of range
```

First, the error when not finding a matching artefact should be better handled.

Second, + should attempt to match a semver version number.

sosna:
Close #592 

This PR does two things:

- Improve error handling by returning a NotFound error in case a matching dataflow is not found in the JSON payload.
- Match stable semver number in case `+` is used as versions.

The tests payloads are the same as the previous ones, except for the modified version number. The drawback is that they are more verbose than what we need. The advantage is that we can reuse the exact same set of tests, regardless of the version number.


javihern98 (approved):
Looks good for me, tests are a bit verbose as you say, otherwise looks great  :blush:

PR #599, opened 2026-05-21
Fix TypeError when SDMX reference URN contains "Ref" substring (#596)

### LINKED ISSUE #596: Concept identies with "Ref" in the name cause a TypeError
When trying to use this library to parse documents with concept identies with identifier including the string "Ref" (such as "ReferentieDatum") a type error is thrown as can be seen below. 

`def __format_component(
        self, comp: Dict[str, Any], role: Role, element_info: Dict[str, Any]
    ) -> Component:
        comp[ROLE.lower()] = role
        comp[REQUIRED] = True
    
        self.__format_local_rep(comp) if LOCAL_REP in comp else None
    
        if LINK in comp:
            del comp[LINK]
        if REF in comp[CON_ID]:
>           concept_id = self.__format_con_id(comp[CON_ID][REF])
                                              ^^^^^^^^^^^^^^^^^
E           TypeError: string indices must be integers, not 'str'

..\.venv\Lib\site-packages\pysdmx\io\xml\__structure_aux_reader.py:829: TypeError
`

I would suggest either making the line ' if REF in comp[CON_ID]: ' more precisely check for a reference concept instead of checking if concept identy contains the string "Ref", or changing REF to a string that includes a special character that indicates the next part is the ref, such as REF = "Ref:" or "Ref." or "Ref="

javihern98:
## Summary
- Resolves #596: SDMX-ML readers raised `TypeError: string indices must be integers, not 'str'` for any reference whose URN contained the literal substring `Ref` (e.g. concept id `ReferentieDatum`).
- Root cause: `if REF in X:` did a key lookup when `X` was a dict but a substring search when `X` was a URN string (SDMX 3.0 shape), mis-routing the string into the dict-subscript branch.
- Fix: guard every such site with `isinstance(X, dict) and REF in X`. Applied at five locations — `ConceptIdentity`, `DataProvider`, `Dataflow` (VTL mapping), `AttachmentGroup`, and the message-header `Structure` reference. The `__format_dataflow` type hint is widened to `Union[str, Dict[str, Any]]` to match the runtime contract.

## Test plan
- [x] New regression test `test_data_structure_concept_id_with_ref` in `tests/io/xml/sdmx30/reader/test_reader.py` parses a fixture with a `ConceptIdentity` URN ending in `ReferentieDatum` — fails on `develop`, passes here.
- [x] `poetry run pytest --cov=pysdmx --cov-branch --cov-fail-under=100 tests/` → 4850 passed, 100 % branch coverage.
- [x] `poetry run ruff format` / `ruff check` / `mypy` clean.


sosna (approved):
Looks good, thanks, @javihern98 

That was an interesting one 😊!

PR #608, opened 2026-05-28
Fix issue with DateTimeFilter not being supported in MultiFilter queries

### LINKED ISSUE #606: DateTimeFilter not always supported in data queries
The following script throws an error.

```python
from pysdmx.api.dc.util import parse_query
from pysdmx.api.qb import ApiVersion, DataContext, DataQuery

qs = "FREQ = 'M' AND REF_AREA = 'RU' AND TIME_PERIOD >= '2018-01'"
flts = parse_query(qs)

dq = DataQuery(DataContext.DATAFLOW, "BIS", "WS_CBPOL", "1.0", components=flts)

url = dq.get_url(ApiVersion.V2_0_0, True)
print(url)
```
The error is: pysdmx.errors.Invalid: Validation Error: Unsupported filter type: DateTimeFilter(field='TIME_PERIOD', operator=<Operator.GREATER_THAN_OR_EQUAL: '>='>, value=datetime.datetime(2018, 1, 28, 0, 0, tzinfo=datetime.timezone.utc))

We need to investigate why this is happening.

sosna:
This PR addresses two issues:

- `DateTimeFilter` was not supported in `DataQuery` when used as part of a `MultiFilter` query.
- The string representing the datetime was not compatible with ISO 8601.

Close #606

### DIFF
--- a/src/pysdmx/api/qb/data.py
+++ b/src/pysdmx/api/qb/data.py
@@ -598,6 +598,12 @@ def __map_operator(op: Operator, value: Any) -> str:
     return out
 
 
+def _format_component_value(value: Any) -> str:
+    if isinstance(value, datetime):
+        return value.isoformat("T", "seconds")
+    return str(value)
+
+
 def _create_component_filter(
     flt: Union[BooleanFilter, DateTimeFilter, NumberFilter, TextFilter],
 ) -> str:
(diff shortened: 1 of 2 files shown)


javihern98 (approved):
Looks good, thanks! :blush:

PR #609, opened 2026-05-28
Fix issue with eager conversion of time periods into datetime objects

### LINKED ISSUE #605: Query parsers use "current" for any missing information in a date time
Query parsers try and parse SDMX time periods as date times. 

However, they wrongly use the current time to complete any missing information. For example, `2026-05` is passed as string, and you run this on the 28th, then the inferred day will be 28, which is not what we want. 

Therefore, in case the date time information is incomplete, we should use a TextFilter, rather than a DateTime filter.

sosna:
Close #605


javihern98 (approved):
Looks good, thanks! :blush:

Issue #612, opened 2026-05-29
Add support for hierarchy formal levels in SDMX-ML


sosna:
We have been requested to add support for hierarchy formal level (cf. #607). While support in the model and SDMX-JSON will be added as part of the PR to address issue #607, it would also be nice to support this in the SDMX-ML readers and writers, for the sake of feature parity between the formats.

PR #613, opened 2026-05-29
Add support for hierarchies with formal levels

### LINKED ISSUE #607: Hierarchy formal levels are silently dropped on read and write.
Hello - We created a curation app that manages an SDMX hierarchy on a remote FMR registry. On every publish, the app's call to `put_structures` silently flipped `hasFormalLevels="true"` to `false` in the registry, breaking every downstream consumption.

If I understand the problem correctly, it looks like hierarchies can declare a chain of formal levels via the `<str:Level>` element (XML) / `level` property (SDMX-JSON 2.0), along with a `hasFormalLevels` flag on the hierarchy itself. These describe the semantic role of each depth in the tree.

pysdmx's `Hierarchy` domain model (`pysdmx.model.code.Hierarchy`) has no level fields at all, and its SDMX-JSON 2.0 mapper (`pysdmx.io.json.sdmxjson2.messages.code.JsonHierarchy`) declares neither `hasFormalLevels` nor `level`. As a result, writing a `Hierarchy` back via RegistryMaintenanceClient .put_structures(...)` produces a JSON body in which  `hasFormalLevels` and `level` are absent. Round-tripping a hierarchy through the registry maintenance API therefore strips its formal levels every time.

pysdmx raises no warning at either end, and neither does FMR.

Thanks!

sosna:
Close #607

This PR adds support for hierarchies with formal levels in:
- The model
- The JSON serializers and deserializers

Support in SDMX-ML is in the scope of another ticket (issue #612).


javihern98 (commented):
@sosna Should we consider moving the DataType and Facets to a different file (maybe enums or representation?) to not over populate the __base.py file. Otherwise looks good :blush:

javihern98 (approved):
Approving this as we can leave the Facets and DataType in __base.py with no risk

PR #610, opened 2026-05-29
Fix issue where representation maps can be created without source or target

sosna:
This PR addresses issue #569 by validating that sources and targets are present when creating a (multi) representation map.

Issue #569, opened 2026-04-23
Creating publication ready artefacts


tonyfujs:
Hello @sosna and @javihern98 - One of our typical workflow using pysdmx is to develop SDMX artefacts locally while iterating on a data pipeline. Once everything looks correct locally, we publish those locally developed artefacts to FMR. A recurring issue is that the locally developed artefacts do not pass validation when pushed to FMR.  

A typical example is missing `urn`. We developed custom functions to handle these cases, but it seems it would be more natural to have stricter validation options in pysdmx.

Let me know what you think, and if something is unclear. Thanks!

sosna:
Thanks for feedback, @tonyfujs. Just a question: We do quite a bit of metadata updates in our registries using pysdmx, and we never have validation issues. 

The `urn` issue sounds even stranger, as it's not mandatory in the FMR, AFAIK (and is even ignored when set, unless I'm mistaken). 

Could I ask which version of the FMR you are using and the types of maintainable artefacts that trigger these validation problems?

 Thank you.

javihern98:
Hi @tonyfujs could you confirm the status for this issue? Could you check the FMR version you are using so we can try to reproduce it?

tonyfujs:
Hi @javihern98 and @sosna - Sorry for not responding earlier. We are currently using FMR 11.25.0 and the issue happened when trying to upload locally built structure maps. I'll try to attach a reproducible example (at least for the local structure map building) by the end of the week. Thanks!

javihern98:
Dear @tonyfujs any updates on this? We are just reviewing any possible bug before the 1.16.0 publication 😊

tonyfujs:
Hi @javihern98 - Just want to give you a heads-up that I did get a chance to look at this this week. I'll prioritize this for early next week. Thanks for your patience!

tonyfujs:
Hi @javihern98 - Sorry for the lack of reactivity on this one. I saw that you have already released version 1.16 (which I have been using to reproduce the issue)... Anyway, here is an example of the issue I was referring to in this ticket. I can create a minimal representation map in python using pysdmx without getting any warnings about missing elements. Once I try to upload it in FMR however, I get an error message.

```
invalid_rm = RepresentationMap(
    id="RM_INVALID",
    name="Invalid Country Map",
    agency="WB.TEST",
    version="1.0",
    maps=[ValueMap(source="BE", target="BEL")]
)

print(invalid_rm)

xml = write_sdmx(invalid_rm,
[...]

PR #614, opened 2026-06-02
Distinguish RepresentationMap and MultiRepresentationMap in SDMX-ML

### LINKED ISSUE #611: SDMX-ML readers do not distinguish between RepresentationMap and MultiRepresentationMap
Hi Javier,

I noticed an issue in the SDMX-ML readers and this issue is the reason why PR #610 is currently failing.

In `pysdmx` model, we distinguish between `RepresentationMap` (a 1-to-1 relationship) and `MultiRepresentationMap` (1-to-n, n-to-1 and n-to-n relationships).

However, the SDMX-ML readers return `RepresentationMap`, regardless of the cardinality.

Could you please address this issue, so as to unblock PR #610? I can also give it a try, but I thought you'd prefer doing it, as it concerns the SDMX-ML readers and writers.

Thanks,

Xavier

javihern98:
## Summary

Closes #611.

The SDMX-ML (XML) readers always built `RepresentationMap` regardless of cardinality, and the writer could only serialize a single source/target. As a result, n-to-1 / 1-to-n / n-to-n maps were mistyped and could not round-trip.

This PR makes the SDMX-ML reader return `MultiRepresentationMap` when a representation map has more than one source **or** more than one target codelist/datatype, and generalises the writer to serialize both types. This mirrors the rule already used by the SDMX-JSON reader (`len(source) != 1 or len(target) != 1`) and the FMR/Fusion-JSON client, which already distinguished the two types — the XML path was the last one lagging behind.

### Changes
- **Reader** (`__structure_aux_reader.py`): new `__build_representation_map` that picks the type by source/target codelist count (mirroring the existing `__build_component_map`); routed at the construction point.
- **Writer** (`__structure_aux_writer.py`): registered `MultiRepresentationMap` in both `STR_DICT_TYPE_LIST_21/30`, switched `__write_scheme` dispatch to an `isinstance` check, and generalised `__write_representation_map` to emit one `Source/Target{Codelist,DataType}` element per entry.
- **Bug fix found in review**: a representation map declaring multiple `SourceDataType`/`TargetDataType` elements previously crashed (`DataType(['String','Integer'])`); the converter is now applied per element.
- **Tests/sample**: `maps.xml` gains matching second codelists; `test_read_maps` asserts `MultiRepresentationMap` with list source/target; new `maps_datatype_multi.xml` + read/round-trip tests for the multi-datatype case; XML↔JSON round-trip covers a multi map.

### SDMX-ML versions
Readers share one `StructureParser` and writers share both dispatch tables, so all versions are covered. Representation maps are a 3.0+ artifact (not valid in the SDMX-ML 2.1 schema), so the round-trip tests cover 3.0 and 3.1.

## Relationship to #610 (branch `cr-569`)
This unblocks #610. The two PRs are complementary:
- **This PR** (#611): reader/writer produce/consume the correct type + io tests + the `maps.xml` second codelists.
- **#610** (#569): model-level validation that source/target counts match the value-map counts.

The `maps.xml` second-codelist additions are intentionally identical to those in `cr-569` to merge cleanly. Consistency between codelist count (container type) and per-mapping value count is enforced by #610's model validation; this PR deliberately does not touch `model/map.py`.

## Verification
- `4857 passed`, **100% branch coverage**
- `ruff format` / `ruff check` clean · `mypy` clean

### DIFF
--- a/src/pysdmx/io/xml/__structure_aux_reader.py
+++ b/src/pysdmx/io/xml/__structure_aux_reader.py
@@ -191,6 +191,7 @@ from pysdmx.model import (
     ImplicitComponentMap,
     KeySet,
     MultiComponentMap,
+    MultiRepresentationMap,
     MultiValueMap,
     RepresentationMap,
     StructureMap,
@@ -240,6 +241,13 @@ T = Any
 
 def _identity(x: T) -> T:
     return x
+
+
+def _convert(converter: Callable[[Any], Any], value: Any) -> Any:
+    """Applies a converter to a value, or to each item if it is a list."""
+    if isinstance(value, list):
+        return [converter(v) for v in value]
+    return converter(value)
 
 
 STRUCTURES_MAPPING = {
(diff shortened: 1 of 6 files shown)


sosna (commented):
Nice, thanks :)

sosna (approved):

PR #621, opened 2026-06-22
Add SDMX-ML support for hierarchies with formal levels

### LINKED ISSUE #612: Add support for hierarchy formal levels in SDMX-ML

javihern98:
Close #612

Extends the formal-levels support added to the model and SDMX-JSON (#607) to the **SDMX-ML readers and writers**. Hierarchies were previously not read or written in any SDMX-ML version, so this adds full `Hierarchy` read+write support (formal levels included) across **2.1, 3.0 and 3.1**, plus `HierarchyAssociation` support (3.0/3.1).

## Changes
- **`Hierarchy` 3.0 / 3.1**: read/write `<str:Hierarchy>` (standalone maintainable) with `hasFormalLevels`, a recursive `<str:Level>`, and `HierarchicalCode`s referencing codes by URN plus an optional per-code level id.
- **`Hierarchy` 2.1**: read/write `<str:HierarchicalCodelist>` wrapping the hierarchy. Reading resolves codes from a `<Code>` `Ref`/`URN` **or** the `CodelistAliasRef` + `CodeID` alias form (via `IncludedCodelist`); writing emits direct `<Code><Ref/></Code>` references.
- **`HierarchyAssociation` 3.0 / 3.1**: read/write `<str:HierarchyAssociation>` (`LinkedHierarchy` ↔ `hierarchy`, `LinkedObject` ↔ `component_ref`, `ContextObject` ↔ `context_ref`). This mirrors the existing model + SDMX-JSON support, which previously had no SDMX-ML counterpart.

## Notes / design decisions
- The 2.1 `Hierarchy` mapping is necessarily lossy (the flat `Hierarchy` model vs. the nested `HierarchicalCodelist`); round-trips are lossless for the common one-hierarchy-per-codelist case.
- `HierarchyAssociation` has no SDMX-ML 2.1 construct (introduced in SDMX 3.0), so it is 3.0/3.1 only.
- Not serialized, matching existing SDMX-JSON/Fusion behaviour (feature parity): `CodingFormat` (level `dtype`/`facets`), `Hierarchy.operator`, and `HierarchyAssociation.operator` (the latter maps to a JSON `Link`, which SDMX-ML I/O does not handle).
- The model is unchanged.

## Verification
- New reader tests (2.1 direct + alias, 3.0, 3.1; hierarchy associations), round-trip tests and writer error-path tests; all sample XML passes `validate=True` (schema-valid).
- `ruff format`, `ruff check`, `mypy` and `pytest` (100% branch coverage) all pass.

### DIFF
--- a/tests/io/xml/sdmx21/writer/test_structures_writing.py
+++ b/tests/io/xml/sdmx21/writer/test_structures_writing.py
@@ -178,7 +178,7 @@ def hierarchy_with_levels():
             name="Division",
             level=LevelType(id="1", name="Group"),
         ),
-        codes=[
+        codes=(
             HierarchicalCode(
                 id="A",
                 urn=(
@@ -186,7 +186,7 @@ def hierarchy_with_levels():
                     "Code=BIS:CL_FREQ(1.0).A"
                 ),
                 level="1",
-                codes=[
+                codes=(
                     HierarchicalCode(
                         id="A1",
                         urn=(
@@ -194,9 +194,9 @@ def hierarchy_with_levels():
                             "Code=BIS:CL_FREQ(1.0).M"
                         ),
                     ),
-                ],
+                ),
             ),
-        ],
+        ),
     )
 
 
@@ -208,19 +208,7 @@ def test_hierarchy_21_round_trip(complete_header, hierarchy_with_levels):
     assert "<str:HierarchicalCodelist " in result
     assert 'leveled="true"' in result
     re_read = read_sdmx(result, validate=True).structures[0]
-    assert isinstance(re_read, Hierarchy)
-    assert re_read.id == "H1"
-    assert re_read.agency == "BIS"
-    assert re_read.has_formal_levels is True
-    assert re_read.level.id == "0"
-    assert re_read.level.level.name == "Group"
-    code_a = re_read.codes[0]
-    assert (
-        code_a.urn
-        == "urn:sdmx:org.sdmx.infomodel.codelist.Code=BIS:CL_FREQ(1.0).A"
-    )
-    assert code_a.level == "1"
-    assert code_a.codes[0].id == "A1"
+    assert re_read == hierarchy_with_levels
 
 
 def test_hierarchy_21_no_levels(complete_header):
@@ -229,7 +217,7 @@ def test_hierarchy_21_no_levels(complete_header):
         name="Flat hierarchy",
         agency="BIS",
         version="1.0",
-        codes=[
+        codes=(
             HierarchicalCode(
                 id="A",
                 urn=(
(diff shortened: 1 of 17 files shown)


tests/io/xml/sdmx21/writer/test_structures_writing.py
```diff
@@ -162,6 +165,142 @@ def complete_header():
...
+                            "Code=BIS:CL_FREQ(1.0).M"
+                        ),
+                    ),
+                ],
+            ),
+        ],
+    )
+
+
+def test_hierarchy_21_round_trip(complete_header, hierarchy_with_levels):
```
sosna, line 203: In this test, should we not assert that hierarchy_with_levels == re_read?

tests/io/xml/sdmx21/writer/test_structures_writing.py
```diff
@@ -162,6 +165,142 @@ def complete_header():
...
+    assert re_read.has_formal_levels is False
+    assert re_read.level is None
+    assert (
+        re_read.codes[0].urn
+        == "urn:sdmx:org.sdmx.infomodel.codelist.Code=BIS:CL_FREQ(1.0).A"
+    )
+
+
+def test_hierarchy_21_metadata_round_trip(complete_header):
+    hierarchy = Hierarchy(
```
sosna, line 237: Same question as above, especially as we don't seem to check the codes in this test...

tests/io/xml/sdmx31/writer/test_structures_writing.py
```diff
@@ -635,6 +640,174 @@ def test_codelist(complete_header, codelist, codelist_sample):
...
+    result = write(
+        [hierarchy_with_levels], header=complete_header, prettyprint=True
+    )
+    assert "<str:Hierarchies>" in result
+    assert 'hasFormalLevels="true"' in result
+    assert '<str:Level id="0">' in result
+    assert '<str:Level id="1">' in result
+    re_read = read_sdmx(result, validate=True).structures[0]
+    assert isinstance(re_read, Hierarchy)
+    assert re_read.has_formal_levels is True
```
sosna: Same question as above

sosna (commented):
Hi Javier. It looks good, thanks. I just had a question about the tests but if you disagree, let me know and I'll approve.

tests/io/xml/sdmx21/writer/test_structures_writing.py
javihern98: Switched to `assert re_read == hierarchy_with_levels` (kept the XML-substring asserts since those verify the serialization, which `==` doesn't).

tests/io/xml/sdmx21/writer/test_structures_writing.py
javihern98: Done in a4af02e7 — now `assert re_read == hierarchy`, which also checks the codes (built the fixture with tuples; see the note on the first thread).

tests/io/xml/sdmx31/writer/test_structures_writing.py
javihern98: Done in a4af02e7 — `assert re_read == hierarchy_with_levels` here too.

sosna (approved):
Looks great, thanks, @javihern98!

PR #633, opened 2026-07-03
Add `c` filters to availability queries in the "data discovery and retrieval" connectors

### LINKED ISSUE #632: Add c filter to availability queries
The TWG has decided to add support for the c filter to the availability query level 1 API, and this impacts pysdmx connector for data discovery and data retrieval (as well as its related AI skill).

sosna:
The SDMX-TWG decided to add support for `c` filters in the availability query defined in the "data discovery and retrieval" profile.

The purpose of this PR is to add support for this in the two `pysdmx` connectors supporting that profile.

While working on this, an issue was discovered in `src/pysdmx/api/qb/availability.py` (short queries would not include the `c` filters).

Close #632

### DIFF
--- a/src/pysdmx/api/dc/pd.py
+++ b/src/pysdmx/api/dc/pd.py
@@ -11,6 +11,7 @@ __check_data_extra()
 
 import pandas as pd
 
+from pysdmx import errors
 from pysdmx.api.dc import BasicConnector, MaintainableIdentification
 from pysdmx.api.dc.query import BasicFilter
 from pysdmx.api.dc.rest import SdmxConnector
@@ -184,17 +185,10 @@ class PandasConnector(BasicConnector):
             The requested data, if any. Data are returned as Pandas data frame.
         """
         q = prepare_basic_data_query(dataflow, filters)
+        temp_path: Optional[pathlib.Path] = None
         try:
             # Write response in chunks to temporary file
-            with tempfile.NamedTemporaryFile(
-                mode="wb", suffix=".csv", delete=False
-            ) as f:
-                for chunk in self.__client.stream_data(
-                    q, chunk_size=1_048_576
-                ):
-                    f.write(chunk)
-                f.flush()
-                f.close()
+            temp_path = self.__write_temp_csv(q)
 
             # Infer read parameters (exclude SDMX columns, add data types etc.)
             params: dict[str, Any] = {}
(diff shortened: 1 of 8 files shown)


src/pysdmx/api/dc/pd.py
```diff
@@ -240,7 +245,32 @@ def data(
...
+            raise errors.InternalError(
+                "Unexpected I/O issue",
+                (
+                    "An internal I/O error occurred while creating a "
+                    "temporary file to process SDMX-CSV data."
+                ),
+                {
+                    "original_exception": str(error),
+                },
+            ) from error
```
javihern98, line 287: I believe we can improve the hygiene by unlinking the file if a BaseException is raised (i.e., a Ctrl+C or connection interrupted will not generate a temp file in this case and will not increase the memory usage). Could be beneficial if we are looking into big data files.

```python
def __write_temp_csv(self, query: Any) -> pathlib.Path:
    try:
        with tempfile.NamedTemporaryFile(
            mode="wb", suffix=".csv", delete=False
        ) as f:
            temp_path = pathlib.Path(f.name)
            try:
                for chunk in self.__client.stream_data(
                    query, chunk_size=1_048_576
                ):
                    f.write(chunk)
                f.flush()
                return temp_path
            except BaseException:
                # File is delete=False and its path never reaches the
                # caller's finally; drop it here. BaseException so a
                # Ctrl-C mid-download also cleans up; raise re-propagates.
                temp_path.unlink(missing_ok=True)
                raise
    except OSError as error:
        raise errors.InternalError(
            "Unexpected I/O issue",
            (
                "An internal I/O error occurred while creating a "
                "temporary file to process SDMX-CSV data."
            ),
            {
                "original_exception": str(error),
            },
        ) from error
```
A regression test could mock a 404 on the data endpoint and assert no *.csv is left behind.

javihern98 (commented):
Minor hygiene concern, otherwise code looks fine 😊

src/pysdmx/api/dc/pd.py
sosna: Thanks a lot, @javihern98, this is a good catch :).

I think we have 2 requirements:
- Mapping any non-pysdmx error to a pysdmx one
- Cleaning up the file system.

So, I'll send a proposal for this, that will be slightly different from your code above, but will keep its spirit :). Thanks again!

sosna:
@javihern98, I have tried to address your comments in the new commits. Let me know if I overlooked anything. Thanks!

src/pysdmx/api/dc/pd.py
javihern98: Looks much better, thanks! I will now approve it

javihern98 (approved):
Looks good now after latest changes 😊

PR #634, opened 2026-07-09
Add .Stat Suite connector to retrieve and upload data and structures

### LINKED ISSUE #450: Connector to .Stat Suite to retrieve and upload data and structures
# Overview

This issue will track the R&D efforts to generate a new connector to the .Stat Suite, reflecting here how the API in this product works to upload data (in SDMX-CSV and/or SDMX-ML) and structures (in SDMX-JSON and/or SDMX-ML).

To ensure sending the pysdmx objects to the .Stat Suite, we will use the write_sdmx method with the desired format and send the artefacts to the .Stat Suite. The connector must handle:

- The required authentication
- The host, port and connection details, including PEM files

Related issues can be generated once the general design is approved by both BIS and MeaningfulData.

Contributions are welcome for anyone interested in the [discussion panel](https://github.com/bis-med-it/pysdmx/discussions/451) . This issue is a placeholder for future reference.

albertohernandez1995:
## Summary

Reference implementation for #450 — a connector for the **.Stat Suite** platform that both **retrieves** and **uploads** data and structures. Follows the approach outlined in the issue: pysdmx objects are serialised with `write_sdmx` in the desired format and sent to the target service.

Since #450 is an R&D/design placeholder this is offered as the proposed design and working implementation rather than a closing change.

## What's included

New module `pysdmx.api.stat`:

- **`StatConnector`** (read) — `fetch_dataflow`, `fetch_schema`, `fetch_dataset` against the SDMX-REST **v2** API. Anonymous by default; optional bearer `token` for access-controlled deployments. Known public endpoints via the `StatEndpoints` enum (OECD, ILO, ABS, Pacific, STATEC).
- **`StatUploader`** (write, `@experimental`) — `submit_structure`, `submit_data`, `submit` (structure→data), `delete_data`, `delete_structure`, and `submission_status` (with optional `wait=` polling to a terminal state). Returns typed, immutable results (`SubmissionResult`, `StructureSubmissionResult`) and detects the in-body (HTTP-200) failures the .Stat services report.
- **Auth** — Keycloak bearer tokens: `fetch_token` (password grant) and `refresh_token`, plus browser/federated-token support for IdP-backed deployments.
- **Connection details** — host/port via URLs, `timeout`, and **PEM** trust files on both connectors.

Supporting changes:

- `RestService` gains an optional bearer `token` (`pysdmx.api.qb.service`).
- How-to guide `docs/howto/stat.rst`, runnable `examples/stat_upload_e2e.py`, and a credentials-gated live integration test (`tests/api/stat/test_integration.py`, skipped without `DOTSTAT_TOKEN`).

## Design decisions for review (BIS)

The issue permits formats *"SDMX-CSV and/or SDMX-ML"* (data) and *"SDMX-JSON and/or SDMX-ML"* (structures). This implementation picks one of each, matching what .Stat accepts in practice:

1. **Data → SDMX-CSV 2.0; structures → SDMX-ML 2.1.** Confirm whether ML data upload and/or JSON structure upload are also wanted (→ follow-up issues).
2. **Formats are fixed, not caller-selectable.** The issue's "desired format" wording could imply configurability; currently they're hard-coded to the working formats. Confirm this is acceptable or should be exposed.
3. **`StatUploader` is `@experimental`** — not covered by semver guarantees while the write API stabilises.

## How it was tested

- Full unit suite with `respx`-mocked HTTP: **100% branch coverage**, `ruff` + `mypy` clean.
- **Live-verified** end-to-end against the SIS-CC demo: author structure → submit → upload data → poll to `Success` → download-and-compare → delete data/constraint/dataflow/DSD/concept-scheme → confirm removed.


src/pysdmx/io/xml/__structure_aux_reader.py
```diff
@@ -1273,14 +1277,11 @@ def __format_key_set(self, keyset_elem: Dict[str, Any]) -> KeySet:
...
-                raise NotImplementedError(
-                    "DataConstraint with role='Actual' is not supported, "
-                    "pysdmx only supports maintainable (Allowed) constraints."
-                )
-            del element["role"]
+        # 'role' (SDMX 3.0) and 'type' (SDMX 2.1) carry the Actual/Allowed
+        # distinction, which the simplified model does not capture; drop them
+        # so a constraint of either kind reads as a plain DataConstraint.
+        element.pop("role", None)
+        element.pop(TYPE, None)
```
javihern98, line 1454: This will need to be removed before merge as #635 already will handle this

src/pysdmx/io/xml/__structure_aux_reader.py
```diff
@@ -1238,6 +1238,10 @@ def __format_cube_region(self, region_elem: Dict[str, Any]) -> CubeRegion:
         if KEY_VALUE in region_elem:
             kv_list = add_list(region_elem[KEY_VALUE])
             for kv in kv_list:
+                if VALUE not in kv:
+                    # A KeyValue may instead carry a TimeRange or numeric
+                    # Range, which the simplified model does not represent.
+                    continue
```
javihern98, line 1417: This will need to be removed before merge as #635 already will handle this

tests/api/qb/service/test_service_structure.py
```diff
@@ -121,3 +121,20 @@ def test_called_as_expected(
...
+        structure_format=StructureFormat.SDMX_ML_2_1,
+        token=bearer,
+    )
+    route = respx_mock.get(url__startswith="https://auth.test").mock(
+        return_value=httpx.Response(200, content=b"<x/>")
+    )
+    svc.structure(StructureQuery(StructureType.DATAFLOW, "A", "DF", "1.0"))
+    assert (
+        route.calls.last.request.headers["Authorization"] == f"Bearer {bearer}"
+    )
```
javihern98, line 140: I believe this test does not add anything meaningful

src/pysdmx/io/xml/__structure_aux_reader.py
```diff
@@ -2129,7 +2130,7 @@ def __format_schema(  # noqa: C901
             elif self.is_sdmx_30 and VERSION in element:
                 element[IS_FINAL_LOW] = is_final(element[VERSION])
 
-            if item == DFW:
+            if item == DFW and STRUCTURE in element:
                 if isinstance(element[STRUCTURE], str):
                     ref_obj = parse_urn(element[STRUCTURE])
                     reference_str = (
```
javihern98: This fix should be a separate issue, adding the relevant tests for this there

src/pysdmx/io/xml/header.py
javihern98, line 1: This fix should be a separate issue, adding the relevant tests for this there

tests/api/stat/test_upload.py
```diff
@@ -0,0 +1,508 @@
...
+
+NSI = "https://nsi.test/rest"
+TRANSFER = "https://transfer.test"
+STRUCT_URL = f"{NSI}/rest/structure"
+IMPORT_URL = f"{TRANSFER}/import/sdmxFile"
+STATUS_URL = f"{TRANSFER}/status/request"
+
+_IO_SAMPLES = Path(__file__).parent.parent.parent / "io" / "samples"
+DATA_CSV = _IO_SAMPLES / "data_v1.csv"
+STRUCTURE_XML = _IO_SAMPLES / "dataflow_structure_children.xml"
```
javihern98: No samples nor files should be needed to test the upload. We could generate MaintainableArtefacts and Pandas Dataframes for this

tests/api/stat/test_integration.py
javihern98, line 1: These tests shall be deleted as the examples should just be documentation-based, no additional examples script should be used

examples/fmr_to_stat.py
javihern98, line 1: No new python script must be added, remove this. Only add the documentation tutorials.

examples/stat_fmr_sync.py
javihern98, line 1: No new python script must be added, remove this. Only add the documentation tutorials.

examples/stat_upload_e2e.py
javihern98, line 1: No new python script must be added, remove this. Only add the documentation tutorials.

docs/howto/stat_fmr_sync.rst
```diff
@@ -0,0 +1,113 @@
...
+    )
+
+    def list_md_urns(base, token):
+        svc = RestService(
+            base, ApiVersion.V2_0_0,
+            structure_format=StructureFormat.SDMX_ML_2_1, token=token,
+        )
+        urns = set()
+        for stype in TYPES:
+            q = StructureQuery(stype, "MD", detail=StructureDetail.ALL_STUBS)
```
javihern98: Maybe use an agency_id variable here, always use kwargs in the documentation to show the parameters involved

docs/howto/fmr_to_stat.rst
javihern98, line 1: Ensure you use kwargs in the documentation examples, for better comprehension

docs/howto/stat.rst
```diff
@@ -0,0 +1,211 @@
...
+    narrower type coverage). Override with ``structure_format=`` for a
+    deployment that requires SDMX-ML; a type that cannot be written in
+    the chosen format raises a clear :class:`~pysdmx.errors.Invalid`.
+
+.. note::
+    Submitted (and deleted) content must use an **agency you are
+    authorized to maintain**. On the SIS-CC demo that agency is ``MD``;
+    everything under other agency IDs is read-only. The agency is taken
+    from the artefacts and datasets you pass, so build them under your
+    writable agency.
```
javihern98: This applies only for us, not for a general user of pysdmx

docs/howto/stat.rst
javihern98, line 1: We need to update this file with the authenticacion changes. It must have a clear example on how to use the three authentication methods and how to pass them to the upload connectors

docs/howto/fmr_to_stat.rst
```diff
@@ -0,0 +1,125 @@
...
+
+Downloading with ``references=all`` is deliberate: it returns the whole
+dependency graph and exercises the SDMX-ML readers on it, and it is
+exactly what is forwarded to .Stat.
+
+.. important::
+    Requires the ``pysdmx[data,xml]`` extras, a **writable** FMR, and a
+    **writable** .Stat Suite instance. Everything is created under the
+    agency you are authorised to maintain (``MD`` on the SIS-CC demo).
+
```
javihern98: Remove the MD reference here. The xml extra is really not needed. We need to differentiate always between data and structures, FMR does hold only structures, therefore with only the base package (no extras) this should be possible

docs/howto/fmr_to_stat.rst
```diff
@@ -0,0 +1,125 @@
...
+.. note::
+    The FMR rejects structures whose maintenance agency is unknown, so
+    the target agency must already exist in the ``SDMX:AGENCIES`` scheme.
+    Register it once (e.g. submit an
+    :class:`~pysdmx.model.organisation.AgencyScheme` containing an
+    :class:`~pysdmx.model.Agency` with ``StructureAction.Merge``) before
+    running this tutorial. The SIS-CC demo already has ``MD``.
+
```
javihern98: No need for this note, just mention the Agency must be inside the SDMX:AGENCIES scheme

docs/howto/fmr_to_stat.rst
```diff
@@ -0,0 +1,125 @@
...
+existing ones:
+
+.. code-block:: python
+
+    from pysdmx.api.fmr.maintenance import (
+        RegistryMaintenanceClient,
+        StructureAction,
+    )
+
+    client = RegistryMaintenanceClient(FMR_ENDPOINT, access_token=token)
```
javihern98: User and password could also be used in the RegistryMaintenanceClient, not just the token

src/pysdmx/api/stat/__init__.py
```diff
@@ -0,0 +1,789 @@
...
+    UNESCAP = "https://dataexplorer.unescap.org/"
+    SIMEL_UY = "https://de-mtss.simel.mtss.gub.uy/"
+    STATCAN_CCEI = "https://de-ccei.statcan.gc.ca/"
+    STATCAN_CITH = "https://de-cith.statcan.gc.ca/"
+    ELSTAT = "https://explore.statistics.gr/"
+    SAMOA = "https://data.sbs.gov.ws/"
+    FIJI = "https://data.statsfiji.gov.fj/"
+    BOTSWANA_LMO = "https://de.lmis.hrdc.org.bw/"
+    UGANDA_LMIS = "https://de.lmis.mglsd.go.ug/"
+    SWISS_FSO = "https://stats.swiss/"
```
javihern98: This is not good, given that all of this URLs must point to an API to use the connector, therefore either find the API connector for all of them, or remove them

src/pysdmx/api/stat/__init__.py
```diff
@@ -0,0 +1,789 @@
...
+            pem: Optional PEM file with trusted certificate authorities,
+                for services using a self-signed certificate.
+            timeout: Maximum number of seconds to wait per request.
+        """
+        self._svc = RestService(
+            api_endpoint,
+            ApiVersion.V2_0_0,
+            data_format=DataFormat.SDMX_CSV_1_0_0,
+            structure_format=StructureFormat.SDMX_ML_2_1,
+            timeout=timeout,
```
javihern98, line 282: Given the tests we made, the data format by default should be SDMX-CSV 2.0 and the Structure format must be SDMX-JSON 2.0

src/pysdmx/api/stat/__init__.py
```diff
@@ -0,0 +1,789 @@
...
+        """
+        query = StructureQuery(
+            StructureType.DATAFLOW,
+            agency,
+            id,
+            version,
+            detail=StructureDetail.FULL,
+            references=StructureReference.DESCENDANTS,
+        )
+        return self._svc.structure(query)
```
javihern98: This is only able to download the Dataflows, we actually need something to download ANY maintainable artefact supported by the .Stat Suite

javihern98 (commented):
Need to review also the StatUploader and endpoints defined

tests/api/qb/service/test_service_structure.py
albertohernandez1995: It's the only coverage for authenticated SDMX-REST reads, so I will keep it

tests/api/stat/test_upload.py
albertohernandez1995: Done

tests/api/stat/test_integration.py
albertohernandez1995: Done

examples/fmr_to_stat.py
albertohernandez1995: Done

examples/stat_fmr_sync.py
albertohernandez1995: Done

examples/stat_upload_e2e.py
albertohernandez1995: Done

docs/howto/stat_fmr_sync.rst
albertohernandez1995: Done

docs/howto/fmr_to_stat.rst
albertohernandez1995: Done

docs/howto/stat.rst
albertohernandez1995: Done

docs/howto/stat.rst
albertohernandez1995: Done

docs/howto/fmr_to_stat.rst
albertohernandez1995: Done

docs/howto/fmr_to_stat.rst
albertohernandez1995: Done

docs/howto/fmr_to_stat.rst
albertohernandez1995: Done

src/pysdmx/api/stat/__init__.py
albertohernandez1995: Done

src/pysdmx/api/stat/__init__.py
albertohernandez1995: Done

src/pysdmx/api/stat/__init__.py
albertohernandez1995: Done

tests/api/qb/service/test_service_structure.py
javihern98: Ok, let´s keep it then because I did not notice we added here the token argument to the base RestService

src/pysdmx/api/stat/__init__.py
javihern98: I cannot see any change here, the data_format and structure_format are still unchanged

src/pysdmx/api/stat/__init__.py
javihern98: Still to implement, the goal is to have something similar as the FMR connector

src/pysdmx/io/xml/__structure_aux_reader.py
albertohernandez1995: https://github.com/bis-med-it/pysdmx/issues/651 solved here.

src/pysdmx/io/xml/header.py
albertohernandez1995: https://github.com/bis-med-it/pysdmx/issues/653 solved here

src/pysdmx/api/stat/__init__.py
javihern98: Need to review these endpoints again for the actual API links, if not found, remove them

albertohernandez1995: Added in last commit.

src/pysdmx/io/xml/__structure_aux_reader.py
albertohernandez1995: PR stays open, when it is merged, this will be resolved.

src/pysdmx/io/xml/__structure_aux_reader.py
albertohernandez1995: PR stays open, when it is merged, this will be resolved.

src/pysdmx/api/stat/__init__.py
albertohernandez1995: It is done

src/pysdmx/api/stat/__init__.py
albertohernandez1995: Keeping SDMX-ML 2.1 + SDMX-CSV 1.0 deliberately — verified live against the
default endpoint (OECD, `sdmx.oecd.org/public/rest/v2`):

- Structures: `Accept: sdmx.structure+json;version=2.0.0` → **406 Not Acceptable**;
  only SDMX-ML 2.1 returns 200. SDMX-ML 2.1 is served by all deployments in the
  `StatEndpoints` enum, so it's the portable choice.
- Data: OECD's SDMX-CSV 2.0 puts `DATAFLOW` (uppercase) in the STRUCTURE column,
  which our own SDMX-CSV 2.0 reader rejects (`Invalid value on STRUCTURE column`).
  SDMX-CSV 1.0 parses fine (41 rows in a sample query).

The connector parses via `read_sdmx` (auto-detection), so these settings only
drive the `Accept` header — the model objects returned are identical to the
JSON/CSV-2.0 path. Unlike `RegistryClient` (which targets the FMR, serving
SDMX-JSON), this client targets .Stat deployments that serve SDMX-ML + SDMX-CSV,
so the defaults necessarily differ.

(The OECD CSV-2.0 uppercase-`DATAFLOW` behaviour looks like a spec-conformance
issue worth a separate ticket — either OECD's output or our reader's strictness —
but until it's resolved, CSV 1.0 is the working default.)

javihern98:
While testing the connector against the SIS-CC demo (`nsi-demo-stable.siscc.org`), I hit an edge case: querying an artefact that does not exist (e.g. right after deleting it) makes `StatConnector.get_dataflows("MD", "DF_DUMMY", "1.0")` raise `Invalid: Cannot parse input as SDMX` instead of raising `NotFound` / returning an empty list.

Cause: .Stat's SDMX-REST **v2** route answers a no-match structure query with `204 No Content` and an **empty body**, unlike the v1 route which returns 404:

```text
GET /rest/dataflow/MD/DF_DUMMY/1.0               -> 404 "No Results Found"
GET /rest/v2/structure/dataflow/MD/DF_DUMMY/1.0  -> 204 (empty body)
```

Since 204 passes `raise_for_status()`, `RestService.structure()` hands `read_sdmx` zero bytes, which fails to parse inside `StatConnector._structures`.

Suggestion: short-circuit `_structures` (sync and async) to return `()` when the raw response is empty — `_one` then raises its `NotFound` and `_many` returns `[]`, which is the expected behaviour for both.

javihern98:
Second finding from live testing against the SIS-CC demo (follow-up to the 204 comment):

**A data import that completes with warnings reports `success=False`.** An import of 4 observations finished with `executionStatus: Completed`, `outcome: Warning`; the data landed and is served back correctly by the NSI. But `_submission_from_status` computes `success = (outcome == "Success")`, so the returned `SubmissionResult.success` is `False` and callers cannot tell a completed-with-warnings import apart from a real failure without digging into `execution_status`/`outcome`/`logs`.

In my opinion a completed import with warnings and no error should be `success=True` — e.g. `success = outcome in {"Success", "Warning"}` — while keeping the raw `outcome`/`logs` fields for callers that want to be strict.

Two related notes from the same test:
- The warning itself looks NSI-side: the SDMX-JSON structure submission declared `localRepresentation.format.dataType: "Double"` for the measure, yet the stored DSD's `PrimaryMeasure` came back with no representation at all, which triggered the *"no data type specified for the measure"* warning on import. Submitting the structures as SDMX-ML 2.1 avoids it.
- `submit()` does not forward a `structure_format` to `submit_structure`, so the combined structure+data flow can only use the SDMX-JSON default; a pass-through parameter would help with deployments like this one.

javihern98:
Third finding from live testing against the SIS-CC demo: **structure-submission rejections reach the caller through two different mechanisms**, depending on how the NSI chooses to answer — and callers cannot predict which one they will get.

- **Logical failure inside an HTTP 2xx** — the case `submit_structure` is designed around: the body is a `SubmitStructureResponse` with per-artefact codes/messages, `_structure_result` parses it, and the caller gets `StructureSubmissionResult(success=False, messages=(...))`. This is what the docstring documents.
- **HTTP-level rejection (422) with an SDMX-ML `message:Error` body** — this path never reaches `_structure_result`: `raise_for_status()` fires inside `_send`, `map_httpx_errors` raises `errors.Invalid`, and the per-artefact texts end up **unparsed, embedded as raw XML in the exception message**. So the same kind of information (per-artefact "no") is surfaced either as a returned result or as a raised exception, in different shapes.

### Steps to reproduce (writable .Stat, e.g. the SIS-CC demo NSI)

1. Author a minimal graph under a maintainable agency: ConceptScheme + DSD (`REF_AREA` dimension, `TIME_PERIOD` time dimension, `OBS_VALUE` measure) + Dataflow, version 1.0.
2. `uploader.submit_structure(structures)` with the SDMX-JSON default → accepted (note: the NSI drops the measure's `localRepresentation` from the JSON, see the previous comment).
3. `uploader.submit_data(dataset)` with a few observations and wait for `Completed` — this also auto-creates the `CR_A_<dataflow>` constraint.
4. Re-submit the same graph with the measure's data type changed (e.g. now `local_dtype=DataType.DOUBLE`, submitted as `structure_format=Format.STRUCTURE_SDMX_ML_2_1`).
5. The NSI answers **HTTP 422** with three `ErrorMessage code="150"` elements, and `submit_structure` **raises** `errors.Invalid` instead of returning `success=False`:

```text
pysdmx.errors.Invalid: Client error 422: ... The error message was: `<?xml version="1.0" encoding="utf-8"?>
<message:Error ...>
  <message:ErrorMessage code="150">
    <common:Text xml:lang="en">Not Updated: Data Structure Definition MD:DSD_DUMMY (v1.0).
    Warning: ... not updated due to changed type: Primary Measure MeasureDescriptor.OBS_VALUE ...
    Other content changes cannot be done until the following dependent artefacts have been deleted:
    Dataflow MD:DF_DUMMY (v1.0)</common:Text>
  </message:ErrorMessage>
  ...
```

### Suggestion

`_structure_result`'s regex (`code="(\d+)"[^>]*>\s*<[^>]*Text[^>]*>(.*?)</`) already matches this 422 body (`code="150"` + `<common:Text>`), so unifying is cheap: catch the 4xx in `submit_structure` (or don't raise for it in `_send` for this call), run the body through the same `_structure_result`, and return `StructureSubmissionResult(success=False, messages=(...))` — keeping exceptions for auth/transport/server errors only. Alternatively, if the raise-on-4xx split is intentional, the docstring should state that semantic rejections can also arrive as `errors.Invalid`, not only as `success=False` (and ideally the raised error would carry the parsed messages).

javihern98:
Fourth finding, completing the deletion story: **the Transfer service keeps per-DSD data-side objects that structure deletion does not remove — and `StatUploader` has no wrapper for the cleanup method that drops them.**

Continuing from the previous comment's scenario: after a full teardown (`delete_data` → constraint → dataflow → DSD → concept scheme, all reported deleted by the NSI), we re-uploaded the graph as SDMX-ML 2.1 with the measure's type changed (String → Double). The structures were inserted fine, but the data import failed with `Completed / outcome: Error`:

> A non-backward-compatible change has been detected in the data structure MD:DSD_DUMMY(1.0) or in its underlying components (...) for which data has been loaded previously (...). When changing a data structure or its components, all corresponding data database objects need to be cleaned and re-created. For that, please delete the concerned data structure(s) and run the Transfer web service's **/cleanup/dsd** method (...). The detected change was: The textformat representation (possible values) of the measure MD:DSD_DUMMY(1.0).OBS_VALUE has changed from String to Double.

The remedy works. Per the Transfer swagger the method is `DELETE {transfer}/{v}/cleanup/dsd`, `multipart/form-data` with required `dataspace` and `dsd` (e.g. `dsd=MD:DSD_DUMMY(1.0)`; optional `cleanupMsdOnly`). After calling it, the identical upload + import completed with outcome `Success`, and the stored `PrimaryMeasure` finally retained `<TextFormat textType="Double"/>` — also confirming that the SDMX-ML 2.1 path preserves the measure representation that SDMX-JSON ingestion drops (see the earlier comment).

Suggestion: add a wrapper, e.g. `StatUploader.cleanup_dsd(dsd, dataspace=None)`, alongside `delete_data`/`delete_structure`. Without it, an uploader that changes a component type is wedged with no pysdmx-native way out (we had to go through the private `_send`). A line in the deletion docs would also help: fully removing a dataflow whose structure will be re-created with changes means `delete_data` → constraint → dataflow → DSD (→ concept scheme) → `cleanup/dsd`.

docs/howto/stat.rst
```diff
@@ -0,0 +1,248 @@
...
+    and :class:`~pysdmx.api.stat.authentication.KeycloakAuthentication`
+    (resource-owner password grant, for local Keycloak accounts only).
+    Alternatively, pass any bearer token obtained elsewhere (for example
+    the Transfer service's Swagger "Authorize" button) directly as
+    ``token=``.
+
+.. note::
+    ``submit_structure`` returns a ``StructureSubmissionResult``
+    (``.success`` / ``.messages``) and reports a per-artefact failure in
+    that body rather than raising. ``submit`` is the convenience wrapper
```
javihern98: Tested live: this does not always hold. Semantic rejections can also arrive as an **HTTP 422** with a `message:Error` body (`code="150"` "Not Updated", e.g. when re-submitting a DSD whose component type changed while dependents exist) — in that case `submit_structure` **raises** `errors.Invalid` (raw XML embedded in the message) and nothing is returned. Suggest noting that rejections can surface both ways, or unifying the channels in code (see the 422 comment in this thread — `_structure_result`'s regex already parses that body).

docs/howto/stat.rst
```diff
@@ -0,0 +1,248 @@
...
+    that submits the structures and then the data; it raises
+    :class:`~pysdmx.errors.Invalid` if the structure step does not
+    succeed.
+
+.. note::
+    ``submit_structure`` serializes to **SDMX-JSON 2.0 by default** --
```
javihern98, line 208: Caveat from live testing: at least one deployment (the SIS-CC demo NSI) accepts SDMX-JSON but silently **drops the measure's local representation** during ingestion — the stored `PrimaryMeasure` had no `TextFormat`, imports warned "no data type specified for the measure", and correcting the type later required full teardown plus the Transfer `/cleanup/dsd`. Suggest advising to verify the stored artefacts after a JSON submission, or use `STRUCTURE_SDMX_ML_2_1` for .Stat targets (needs the `xml` extra).

docs/howto/stat.rst
```diff
@@ -0,0 +1,248 @@
...
+observations by uploading them with the SDMX *Delete* action, and delete
+structures with :meth:`~pysdmx.api.stat.StatUploader.delete_structure`.
+
+.. code-block:: python
+
+    # Delete observations (SDMX-CSV ACTION=D under the hood)
+    uploader.delete_data(dataset)
+
+    # Delete structures. A data import auto-creates an actual content
+    # constraint (CR_A_<dataflow>); tear down in dependency order:
```
javihern98, line 254: From live testing: this teardown is not sufficient when the DSD will be re-created **with changed component types**. The Transfer service keeps per-DSD data-side objects that NSI deletion does not remove, and the next import fails with "non-backward-compatible change … String to Double" until `DELETE {transfer}/{v}/cleanup/dsd` (form fields `dataspace`, `dsd="AG:ID(ver)"`) is called. Worth a note here — or the `cleanup_dsd()` wrapper suggested earlier in the thread, with the docs following.

docs/howto/stat.rst
```diff
@@ -0,0 +1,248 @@
...
+        "DataStructure=MD:DSD_EXAMPLE(1.0)",
+        "ConceptScheme=MD:CS_EXAMPLE(1.0)",
+    ])
+
+``delete_structure`` accepts maintainable artefacts or short-URN strings,
+and deletes them in the order given — pass dependents first, or a
+still-referenced artefact yields :class:`~pysdmx.errors.Invalid` (HTTP
+409). It returns one ``StructureSubmissionResult`` per artefact (check
+each ``.success``). Like ``submit_structure``, both report a logical
+(in-body) failure through ``.success`` / ``.messages`` rather than
```
javihern98: This contradicts `delete_structure`'s own docstring ("The NSI signals a delete failure with an HTTP status … rather than an in-body error") and what we observed live: delete failures (404/409) **raise** `NotFound`/`Invalid`, fail-fast. Only successful deletions produce parsed results. Suggest rewording: for deletions the returned results are informational, and failures surface as exceptions.

docs/howto/stat.rst
```diff
@@ -0,0 +1,248 @@
...
+    result = uploader.submit(structures=dataflow, dataset=dataset)
+
+    # Poll to completion, then check the outcome
+    final = uploader.submission_status(result.request_id, wait=True)
+    print(final.execution_status, final.outcome)
+
+``submit``/``submit_data`` return a ``SubmissionResult`` whose
+``request_id`` identifies the asynchronous transaction; feed it to
+``submission_status`` to obtain the final ``execution_status`` and
+``outcome``. Pass ``wait=True`` to poll until the transaction reaches a
```
javihern98, line 164: Note from live testing: an import that lands with warnings finishes `executionStatus: Completed` / `outcome: "Warning"` — and `SubmissionResult.success` is then `False` (only `outcome == "Success"` counts). Depending on the decision on the Warning-semantics comment in this thread, either the code or this paragraph should make that explicit, so callers do not treat completed-with-warnings as a failure.

docs/howto/stat_fmr_sync.rst
```diff
@@ -0,0 +1,130 @@
...
+        dataspace=SPACE,
+        token=dotstat_token,
+    )
+    for result in uploader.delete_structure(ordered):
+        state = "ok" if result.success else "FAILED"
+        print(state, "; ".join(result.messages))
+
+``delete_structure`` returns one
+:class:`~pysdmx.api.stat.StructureSubmissionResult` per artefact, so a
+logical failure surfaces as ``success=False`` rather than an exception.
```
javihern98: As written, this loop cannot report a failure: `delete_structure` is **fail-fast** — on the first NSI error (404/409) it raises `NotFound`/`Invalid` mid-sequence (with an indeterminate prefix already deleted), so the `FAILED` branch never runs. Verified live: delete failures arrive as HTTP statuses, not in-body results. Suggest wrapping the call in `try/except` (and re-running with the remaining URNs) and correcting the sentence below the snippet.

docs/howto/fmr_to_stat.rst
```diff
@@ -0,0 +1,136 @@
...
+3. download it back from the FMR **with all its references**;
+4. submit the downloaded structures and a tiny dataset to .Stat.
+
+Downloading with ``references=all`` is deliberate: it returns the whole
+dependency graph, which is exactly what is forwarded to .Stat.
+
+.. important::
+    Requires a **writable** FMR and a **writable** .Stat Suite instance.
+    Moving *structures* needs only the base ``pysdmx`` package (they are
+    exchanged as SDMX-JSON); the ``data`` extra (``pysdmx[data]``) is
```
javihern98, line 26: Caveat worth adding here (verified live): .Stat's SDMX-JSON ingestion can drop component representations (e.g. the measure's data type — see the thread), so the JSON-only path may store a degraded DSD on the .Stat side. If representations matter, submit to .Stat as SDMX-ML 2.1 (`xml` extra) or verify the stored structures after upload.

javihern98 (commented):
Fifth item in the series: the doc pages vs. the live test cycle documented in this thread.

Cross-checked the four .Stat doc pages against the full lifecycle we ran on the SIS-CC demo (upload → import → type change → 422 → teardown → `/cleanup/dsd` → clean re-upload → verified reads). **Confirmed by testing and fine as written**: the endpoint forms (NSI host + `/rest/v2` for reads, Transfer including its version segment), realm-URL-only device auth, the async import flow (`request_id` → `submission_status(wait=True)` → `execution_status`/`outcome`), actions carried in the file, the auto-created `CR_A_<dataflow>` constraint with dependents-first deletion order, and short-URN strings in `delete_structure`.

The inline comments mark the places that need updating — one is contradicted by the library's own behaviour, the rest are missing caveats we hit in practice.

javihern98:
Sixth item: results of an **error battery** against the SIS-CC demo — six submission scenarios probing how failures surface through the connector (graph: Codelist + ConceptScheme + DSD with an enumerated dimension + Dataflow, under MD; SDMX-ML 2.1; demo cleaned before and after).

**1. A third rejection channel: HTTP 500 for client-side mistakes.** Submitting a dataflow alone, referencing a DSD that does not exist, returns HTTP 500 with a proper SDMX error body — surfaced by pysdmx as `errors.InternalError`:

> `<message:ErrorMessage code="500">` Dataflow DF_DUMMY uses the DSD: ID: DSD_DUMMY VERSION:1.0 AGENCY: MD which doesn't exist in the Mapping Store

So alongside the in-body 2xx results and the 422 envelope (earlier comments), a dangling reference lands in the `InternalError` bucket, which callers reasonably interpret as a *server* fault. Retry/error-handling logic keyed on pysdmx's hierarchy will misclassify these.

**2. Structure submissions are per-artefact, not atomic.** Submitting ConceptScheme + DSD where the DSD references an absent codelist gives `success=False` — but the message shows the CS **was inserted** while only the DSD was rejected:

> Created: Concept Scheme MD:CS_DUMMY (v1.0). Success (…) was inserted.
> Not Created: Data Structure Definition MD:DSD_DUMMY (v1.0), Missing references - Codelist MD:CL_AREA (v1.0)

`StructureSubmissionResult.success=False` therefore does **not** mean nothing changed — worth stating in the docs (and it strengthens the case for keeping the per-artefact `messages` accessible on every rejection channel).

**3. Unauthorized agency → HTTP 401, and pysdmx blames the token.** Submitting a codelist under agency `ZZ9` (not in `SDMX:AGENCIES`) returns 401, which `_send` reports as *"The service rejected the token"*. The token was valid — .Stat uses 401 for content authorization. Suggest wording like "the token was rejected **or lacks permission for the submitted content/agency**".

**4. No idempotent re-submit.** Re-sending the *identical* graph right after a successful insert (no data loaded) fails with HTTP 500 and a plain-text `Internal server error` body (no SDMX envelope). With data/dependents present, the same action fails as 422 code 150 (earlier comment). Either way there is no safe "ensure structures exist" upsert against .Stat — publishers must expect an error on unchanged re-submission, in a state-dependent shape.

**5. `delete_structure`'s `NotFound` never fires on this NSI.** During cleanup, deleting `Codelist=MD:CL_AREA(1.0)` **before it had ever been submitted** returned 2xx — DELETE is idempotent for missing artefacts under an authorized agency, so the docstring's "Raises NotFound: If an artefact does not exist" did not hold on this deployment (only the unauthorized-agency delete errored, with 401).

Positive results from the same battery: the full graph including the codelist inserts cleanly as SDMX-ML 2.1 (enumeration reference intact), and the Transfer service enforces enumerations on import with row-precise diagnostics ("The value for dimension REF_AREA:IT at observation 2 (… row 3) is incorrect") — noting that one bad row drops the **whole batch**, including valid rows.

javihern98:
Seventh (small) item: **data bound to the DSD is refused — imports must reference a Dataflow.**

Tested on the SIS-CC demo: submitting a dataset whose schema comes from `dsd.to_schema()` (so the SDMX-CSV carries `STRUCTURE=datastructure, STRUCTURE_ID=MD:DSD_DUMMY(1.0)`) is registered by the Transfer service and then dropped, with a clean and explicit message:

> Importing data is not supported when the referenced structure is a DataStructure. Please change the referenced structure to a DataFlow and re-submit the request.

Nothing is written (verified by read-back). The pysdmx angle: the library naturally produces such datasets (`DataStructureDefinition.to_schema()` → context `"datastructure"`), and `submit_data` serializes and uploads them without complaint — the user only finds out at status-poll time.

Suggestion: add a note to the data-upload documentation — the `submit_data` docstring (the "must be Schema-backed" paragraph) and the how-to's Uploading section — stating that the dataset's schema must reference a **dataflow** (`context="dataflow"`); a `datastructure`-bound dataset is rejected by .Stat.

src/pysdmx/api/stat/__init__.py
javihern98: This issue with the values in the STRUCTURE column will be checked in https://github.com/bis-med-it/pysdmx/issues/667

javihern98: After some analysis seems that the .Stat Suite only bound the data to a Dataflow, not a DataStructure. Therefore the use of SDMX-CSV 2.0 will only give the ACTION. In any case I prefer to have the latest version by default, consequently let´s change the default to 2.0 as the issue described in the comment above will be solved

albertohernandez1995:
> While testing the connector against the SIS-CC demo (`nsi-demo-stable.siscc.org`), I hit an edge case: querying an artefact that does not exist (e.g. right after deleting it) makes `StatConnector.get_dataflows("MD", "DF_DUMMY", "1.0")` raise `Invalid: Cannot parse input as SDMX` instead of raising `NotFound` / returning an empty list.
> 
> Cause: .Stat's SDMX-REST **v2** route answers a no-match structure query with `204 No Content` and an **empty body**, unlike the v1 route which returns 404:
> 
> ```
> GET /rest/dataflow/MD/DF_DUMMY/1.0               -> 404 "No Results Found"
> GET /rest/v2/structure/dataflow/MD/DF_DUMMY/1.0  -> 204 (empty body)
> ```
> 
> Since 204 passes `raise_for_status()`, `RestService.structure()` hands `read_sdmx` zero bytes, which fails to parse inside `StatConnector._structures`.
> 
> Suggestion: short-circuit `_structures` (sync and async) to return `()` when the raw response is empty — `_one` then raises its `NotFound` and `_many` returns `[]`, which is the expected behaviour for both.

Fixed. `_structures` in both StatConnector and AsyncStatConnector now short-circuits an empty response body: a 204 No Content resolves to NotFound for single-artefact getters (get_codes, …) and to an empty list for list getters (get_dataflows, …), instead of surfacing "Cannot parse input as SDMX". Added sync + async tests.

albertohernandez1995:
> Second finding from live testing against the SIS-CC demo (follow-up to the 204 comment):
> 
> **A data import that completes with warnings reports `success=False`.** An import of 4 observations finished with `executionStatus: Completed`, `outcome: Warning`; the data landed and is served back correctly by the NSI. But `_submission_from_status` computes `success = (outcome == "Success")`, so the returned `SubmissionResult.success` is `False` and callers cannot tell a completed-with-warnings import apart from a real failure without digging into `execution_status`/`outcome`/`logs`.
> 
> In my opinion a completed import with warnings and no error should be `success=True` — e.g. `success = outcome in {"Success", "Warning"}` — while keeping the raw `outcome`/`logs` fields for callers that want to be strict.
> 
> Two related notes from the same test:
> 
> * The warning itself looks NSI-side: the SDMX-JSON structure submission declared `localRepresentation.format.dataType: "Double"` for the measure, yet the stored DSD's `PrimaryMeasure` came back with no representation at all, which triggered the _"no data type specified for the measure"_ warning on import. Submitting the structures as SDMX-ML 2.1 avoids it.
> * `submit()` does not forward a `structure_format` to `submit_structure`, so the combined structure+data flow can only use the SDMX-JSON default; a pass-through parameter would help with deployments like this one.

Done for the warning-outcome part: _submission_from_status now treats outcome="Warning" as a success (in addition to "Success"), and the raw outcome / execution_status fields remain populated so strict callers can still distinguish a completed-with-warnings run. The doc wording (submission_status section) is updated in the docs commit; the other two points raised here — submit() structure_format pass-through and the SDMX-JSON measure-representation caveat — are handled in their own commits.

albertohernandez1995:
> Third finding from live testing against the SIS-CC demo: **structure-submission rejections reach the caller through two different mechanisms**, depending on how the NSI chooses to answer — and callers cannot predict which one they will get.
> 
> * **Logical failure inside an HTTP 2xx** — the case `submit_structure` is designed around: the body is a `SubmitStructureResponse` with per-artefact codes/messages, `_structure_result` parses it, and the caller gets `StructureSubmissionResult(success=False, messages=(...))`. This is what the docstring documents.
> * **HTTP-level rejection (422) with an SDMX-ML `message:Error` body** — this path never reaches `_structure_result`: `raise_for_status()` fires inside `_send`, `map_httpx_errors` raises `errors.Invalid`, and the per-artefact texts end up **unparsed, embedded as raw XML in the exception message**. So the same kind of information (per-artefact "no") is surfaced either as a returned result or as a raised exception, in different shapes.
> 
> ### Steps to reproduce (writable .Stat, e.g. the SIS-CC demo NSI)
> 1. Author a minimal graph under a maintainable agency: ConceptScheme + DSD (`REF_AREA` dimension, `TIME_PERIOD` time dimension, `OBS_VALUE` measure) + Dataflow, version 1.0.
> 2. `uploader.submit_structure(structures)` with the SDMX-JSON default → accepted (note: the NSI drops the measure's `localRepresentation` from the JSON, see the previous comment).
> 3. `uploader.submit_data(dataset)` with a few observations and wait for `Completed` — this also auto-creates the `CR_A_<dataflow>` constraint.
> 4. Re-submit the same graph with the measure's data type changed (e.g. now `local_dtype=DataType.DOUBLE`, submitted as `structure_format=Format.STRUCTURE_SDMX_ML_2_1`).
> 5. The NSI answers **HTTP 422** with three `ErrorMessage code="150"` elements, and `submit_structure` **raises** `errors.Invalid` instead of returning `success=False`:
> 
> ```
> pysdmx.errors.Invalid: Client error 422: ... The error message was: `<?xml version="1.0" encoding="utf-8"?>
> <message:Error ...>
>   <message:ErrorMessage code="150">
>     <common:Text xml:lang="en">Not Updated: Data Structure Definition MD:DSD_DUMMY (v1.0).
>     Warning: ... not updated due to changed type: Primary Measure MeasureDescriptor.OBS_VALUE ...
>     Other content changes cannot be done until the following dependent artefacts have been deleted:
>     Dataflow MD:DF_DUMMY (v1.0)</common:Text>
>   </message:ErrorMessage>
>   ...
> ```
> 
> ### Suggestion
> `_structure_result`'s regex (`code="(\d+)"[^>]*>\s*<[^>]*Text[^>]*>(.*?)</`) already matches this 422 body (`code="150"` + `<common:Text>`), so unifying is cheap: catch the 4xx in `submit_structure` (or don't raise for it in `_send` for this call), run the body through the same `_structure_result`, and return `StructureSubmissionResult(success=False, messages=(...))` — keeping exceptions for auth/transport/server errors only. Alternatively, if the raise-on-4xx split is intentional, the docstring should state that semantic rejections can also arrive as `errors.Invalid`, not only as `success=False` (and ideally the raised error would carry the parsed messages).

Done: submit() now accepts a structure_format parameter (default SDMX-JSON 2.0, same as submit_structure) and forwards it, so a combined structure+data workflow can pick SDMX-ML 2.1 without splitting into submit_structure + submit_data. Added a test asserting the structure POST carries the chosen format's content type.

src/pysdmx/api/stat/__init__.py
albertohernandez1995: Done, now that #668 (case-insensitive STRUCTURE column) is merged: the read connector defaults data_format to SDMX-CSV 2.0 (verified it now parses OECD's output, 41 rows). Structures stay SDMX-ML 2.1 since OECD still returns 406 for SDMX-JSON structures, so that half is unchanged. Updated the in-code rationale comment to match.

docs/howto/stat.rst
albertohernandez1995: Fixed the note. docs/howto/stat.rst and the submit_structure docstring now state that a semantic rejection can also arrive as an HTTP 422 with an SDMX error document (raised as errors.Invalid), not only as the in-body StructureSubmissionResult.success=False path — callers should check both.

docs/howto/stat.rst
albertohernandez1995: Reworded. The Deleting note now states delete_structure is fail-fast: the first artefact the NSI rejects raises errors.Invalid and stops the sequence (an indeterminate prefix may already be deleted), and the returned StructureSubmissionResults are informational — the artefacts deleted before the failure — not an in-body success=False signal.

docs/howto/stat_fmr_sync.rst
albertohernandez1995: Fixed. The delete loop is now wrapped in try/except (Invalid, NotFound) (delete_structure is fail-fast and raises on the first NSI rejection mid-sequence, leaving an indeterminate prefix deleted — re-run with the remaining URNs), and the sentence below the snippet corrected to say a failure raises rather than surfacing as success=False.

albertohernandez1995:
> Seventh (small) item: **data bound to the DSD is refused — imports must reference a Dataflow.**
> 
> Tested on the SIS-CC demo: submitting a dataset whose schema comes from `dsd.to_schema()` (so the SDMX-CSV carries `STRUCTURE=datastructure, STRUCTURE_ID=MD:DSD_DUMMY(1.0)`) is registered by the Transfer service and then dropped, with a clean and explicit message:
> 
> > Importing data is not supported when the referenced structure is a DataStructure. Please change the referenced structure to a DataFlow and re-submit the request.
> 
> Nothing is written (verified by read-back). The pysdmx angle: the library naturally produces such datasets (`DataStructureDefinition.to_schema()` → context `"datastructure"`), and `submit_data` serializes and uploads them without complaint — the user only finds out at status-poll time.
> 
> Suggestion: add a note to the data-upload documentation — the `submit_data` docstring (the "must be Schema-backed" paragraph) and the how-to's Uploading section — stating that the dataset's schema must reference a **dataflow** (`context="dataflow"`); a `datastructure`-bound dataset is rejected by .Stat.

Documented. The submit_data docstring ("must be Schema-backed" paragraph) and the how-to's Uploading section now state the dataset's schema must reference a dataflow (context="dataflow"), not a datastructure — a datastructure-bound dataset (e.g. from DataStructureDefinition.to_schema()) is rejected by .Stat, and the failure surfaces only at status-poll time.

docs/howto/stat.rst
albertohernandez1995: Added the caveat to the SDMX-JSON-default note: some .Stat deployments drop a measure's local representation on SDMX-JSON ingestion (the stored DSD loses the measure's data type/codelist), so submit as SDMX-ML 2.1 or read the artefacts back to verify. (The /cleanup/dsd teardown you mention is documented separately.)

docs/howto/fmr_to_stat.rst
albertohernandez1995: Added a note after the import example: .Stat's SDMX-JSON ingestion can drop component representations (a degraded DSD), so submit as SDMX-ML 2.1 (xml extra) or verify the stored structures after upload.

albertohernandez1995:
> Fourth finding, completing the deletion story: **the Transfer service keeps per-DSD data-side objects that structure deletion does not remove — and `StatUploader` has no wrapper for the cleanup method that drops them.**
> 
> Continuing from the previous comment's scenario: after a full teardown (`delete_data` → constraint → dataflow → DSD → concept scheme, all reported deleted by the NSI), we re-uploaded the graph as SDMX-ML 2.1 with the measure's type changed (String → Double). The structures were inserted fine, but the data import failed with `Completed / outcome: Error`:
> 
> > A non-backward-compatible change has been detected in the data structure MD:DSD_DUMMY(1.0) or in its underlying components (...) for which data has been loaded previously (...). When changing a data structure or its components, all corresponding data database objects need to be cleaned and re-created. For that, please delete the concerned data structure(s) and run the Transfer web service's **/cleanup/dsd** method (...). The detected change was: The textformat representation (possible values) of the measure MD:DSD_DUMMY(1.0).OBS_VALUE has changed from String to Double.
> 
> The remedy works. Per the Transfer swagger the method is `DELETE {transfer}/{v}/cleanup/dsd`, `multipart/form-data` with required `dataspace` and `dsd` (e.g. `dsd=MD:DSD_DUMMY(1.0)`; optional `cleanupMsdOnly`). After calling it, the identical upload + import completed with outcome `Success`, and the stored `PrimaryMeasure` finally retained `<TextFormat textType="Double"/>` — also confirming that the SDMX-ML 2.1 path preserves the measure representation that SDMX-JSON ingestion drops (see the earlier comment).
> 
> Suggestion: add a wrapper, e.g. `StatUploader.cleanup_dsd(dsd, dataspace=None)`, alongside `delete_data`/`delete_structure`. Without it, an uploader that changes a component type is wedged with no pysdmx-native way out (we had to go through the private `_send`). A line in the deletion docs would also help: fully removing a dataflow whose structure will be re-created with changes means `delete_data` → constraint → dataflow → DSD (→ concept scheme) → `cleanup/dsd`.

Added the wrapper: StatUploader.cleanup_dsd(dsd, dataspace=None, cleanup_msd_only=False) issues DELETE {transfer}/cleanup/dsd (multipart/form-data with dataspace + dsd in AG:ID(version) form, and cleanupMsdOnly when requested) and returns a SubmissionResult; it accepts a short URN or a DataStructureDefinition. The how-to Deleting section now documents the full teardown order ending in cleanup_dsd. Implemented to the swagger spec you quoted — not yet live-verified against an NSI on my side, so a quick check on the demo would be worthwhile.

docs/howto/stat.rst
albertohernandez1995: Added the cleanup_dsd() wrapper and referenced it here: the Deleting section now notes NSI deletion is insufficient when a DSD is re-created with changed component types, and calls cleanup_dsd() as the final teardown step (with an example).

albertohernandez1995:
> Sixth item: results of an **error battery** against the SIS-CC demo — six submission scenarios probing how failures surface through the connector (graph: Codelist + ConceptScheme + DSD with an enumerated dimension + Dataflow, under MD; SDMX-ML 2.1; demo cleaned before and after).
> 
> **1. A third rejection channel: HTTP 500 for client-side mistakes.** Submitting a dataflow alone, referencing a DSD that does not exist, returns HTTP 500 with a proper SDMX error body — surfaced by pysdmx as `errors.InternalError`:
> 
> > `<message:ErrorMessage code="500">` Dataflow DF_DUMMY uses the DSD: ID: DSD_DUMMY VERSION:1.0 AGENCY: MD which doesn't exist in the Mapping Store
> 
> So alongside the in-body 2xx results and the 422 envelope (earlier comments), a dangling reference lands in the `InternalError` bucket, which callers reasonably interpret as a _server_ fault. Retry/error-handling logic keyed on pysdmx's hierarchy will misclassify these.
> 
> **2. Structure submissions are per-artefact, not atomic.** Submitting ConceptScheme + DSD where the DSD references an absent codelist gives `success=False` — but the message shows the CS **was inserted** while only the DSD was rejected:
> 
> > Created: Concept Scheme MD:CS_DUMMY (v1.0). Success (…) was inserted.
> > Not Created: Data Structure Definition MD:DSD_DUMMY (v1.0), Missing references - Codelist MD:CL_AREA (v1.0)
> 
> `StructureSubmissionResult.success=False` therefore does **not** mean nothing changed — worth stating in the docs (and it strengthens the case for keeping the per-artefact `messages` accessible on every rejection channel).
> 
> **3. Unauthorized agency → HTTP 401, and pysdmx blames the token.** Submitting a codelist under agency `ZZ9` (not in `SDMX:AGENCIES`) returns 401, which `_send` reports as _"The service rejected the token"_. The token was valid — .Stat uses 401 for content authorization. Suggest wording like "the token was rejected **or lacks permission for the submitted content/agency**".
> 
> **4. No idempotent re-submit.** Re-sending the _identical_ graph right after a successful insert (no data loaded) fails with HTTP 500 and a plain-text `Internal server error` body (no SDMX envelope). With data/dependents present, the same action fails as 422 code 150 (earlier comment). Either way there is no safe "ensure structures exist" upsert against .Stat — publishers must expect an error on unchanged re-submission, in a state-dependent shape.
> 
> **5. `delete_structure`'s `NotFound` never fires on this NSI.** During cleanup, deleting `Codelist=MD:CL_AREA(1.0)` **before it had ever been submitted** returned 2xx — DELETE is idempotent for missing artefacts under an authorized agency, so the docstring's "Raises NotFound: If an artefact does not exist" did not hold on this deployment (only the unauthorized-agency delete errored, with 401).
> 
> Positive results from the same battery: the full graph including the codelist inserts cleanly as SDMX-ML 2.1 (enumeration reference intact), and the Transfer service enforces enumerations on import with row-precise diagnostics ("The value for dimension REF_AREA:IT at observation 2 (… row 3) is incorrect") — noting that one bad row drops the **whole batch**, including valid rows.

Captured the actionable parts:
- (3) Widened the 401 message: token missing/expired OR lacking
  permission for the submitted content/agency (.Stat uses 401 for
  content authorization).
- (2) Documented that structure submission is per-artefact, not atomic
  (success=False can still mean some artefacts were inserted — .messages).
- (5) Adjusted delete_structure's docstring: DELETE is idempotent on many
  NSIs, so NotFound may not fire for an already-absent artefact.
- (1)(4) Added a "known behaviours" note: dangling refs / unchanged
  re-submits can come back as HTTP 500 (no idempotent upsert), and one
  bad observation rejects the whole data batch.
The 500-for-client-errors (1) is the service's own status code, so pysdmx
maps it to InternalError as-is — documented rather than reclassified to
avoid brittle heuristics.

docs/howto/stat.rst
albertohernandez1995: Done on both sides now. The code was changed in this thread (_submission_from_status treats outcome="Warning" as success alongside "Success"), and the submission_status paragraph in the .Stat how-to now states explicitly that a completed-with-warnings import (execution_status="Completed", outcome="Warning") is success=True, with outcome/logs still exposed for callers that want to treat warnings
strictly. So callers no longer see completed-with-warnings as a failure.

src/pysdmx/api/stat/__init__.py
javihern98: Great, thanks!

Issue #561, opened 2026-03-16
AI Coding Assistant Instructions for data analysts


sosna:
`pysdmx` now offers AI Coding Assistant Instructions for developers of the library. However, it would be nice to have something similar for users of the library (for example, on how to use pysdmx to retrieve data from service XYZ).

javihern98:
Hi @sosna, I do not know the state of the art on how could we perform this as of today. I guess a good readme and good documentation as we have today is enough for this. Any ideas?

sosna:
A colleague told me he would send me his own file that he created for this purpose. Let me get back to this when I have received it.

javihern98:
@sosna Happy to start working on this if the approach sounds reasonable. My only concern with this is to build something useful and maintainable over time with minimum effort, I think this proposal with a proper review process covers both 😊 

(Generated by Claude, edited by me for conciseness. Query builder approach to API generation must be discussed beforehand😊 )

## Proposal: AI Agent Documentation for pysdmx

Here is a concrete proposal for how we could structure documentation aimed at AI coding assistants and autonomous agents that want to use pysdmx for data analysis.

### The problem

pysdmx already has great documentation for humans (Sphinx docs, how-to guides, API reference) and AI coding assistant instructions for **developers** of the library (`.claude/CLAUDE.md`). What's missing is documentation optimized for AI agents that want to **use** the library — something that gives them the right mental model, key entry points, and common patterns without having to parse the full API reference.

### Proposed approach: llms.txt convention

We'd follow the [llms.txt convention](https://llmstxt.org/) — a growing standard for providing LLM-friendly documentation alongside human documentation. The files would be **auto-generated** from the existing Sphinx docs using the [`sphinx-llms-txt`](https://pypi.org/project/sphinx-llms-txt/) extension, so they stay in sync with every release automatically.

#### Generated files

| File | Purpose | Consumer |
| ---- | ------- | -------- |
| `llms.txt` (docs site) | Concise overview — project description + links to all doc pages in toctree order | Agents needing quick orientation |
| `llms-full.txt` (docs site) | Full documentation as a single flat text file — all Sphinx pages concatenated in toctree order | Agents needing detailed API knowledge |

Both files are generated during `sphinx-build` and deployed to the docs site alongside the HTML pages. No manual maintenance required.

#### New Sphinx page

[...]

PR #636, opened 2026-07-10
Add AI skill for data discovery and data retrieval

sosna:
Dear all,

The AI skill has been tested by several colleagues now so I think it's good enough to share with a wider community.

We still need to agree on Javier's [generic proposal here](https://github.com/bis-med-it/pysdmx/issues/561), but I guess we can always reorganize things later.

Thanks,

Xavier


javihern98 (approved):
Looks good, thanks! 😊

PR #637, opened 2026-07-10
Support availability constraints and time ranges

### LINKED ISSUE #635: Constraints: align with SDMX 3.1 and support AvailabilityConstraint & TimeRange
## Summary

pysdmx cannot represent two constraint constructs that are common in real-world SDMX messages: the data **actually available** in a data source ("Actual" constraints in SDMX 2.1/3.0) and **time ranges** in cube regions. Since #652 the SDMX-ML readers no longer crash on them, but the information is **silently dropped**: the Allowed/Actual marker is stripped (an "Actual" constraint reads as a plain `DataConstraint`) and a cube-region `KeyValue` holding a `TimeRange` disappears from the parsed model. E.g. reading [this OECD message](https://sdmx.oecd.org/public/rest/dataflow/OECD.SDD.TPS/DSD_SDBSBD_ISIC4@DF_BD_API/1.0?references=all) succeeds, but loses the whole `TIME_PERIOD` coverage (1989→2024) and the fact that the constraint describes actual data. The SDMX-JSON reader likewise ignores `role` and `timeRange`. Availability responses from the REST `availability` endpoint (which pysdmx already queries in `api/qb` / `api/dc`) therefore cannot be fully modelled.

The goal of this issue is to **support the new constraint objects, aligned with SDMX 3.1**, where data availability is a **separate object** (`AvailabilityConstraint`) rather than a role/type flag on the constraint. **`DataConstraint` gets no `role`/`type` property** — the semantics are carried by the class of the object.

## How SDMX 3.1 models availability (and why we follow it)

The Allowed/Actual flag evolved across the standard:

- **SDMX-ML 2.1**: one `ContentConstraint` element with a `type` attribute — optional, **default `Actual`**.
- **SDMX-ML 3.0**: `DataConstraint` with a **required** `role` attribute (`Allowed`/`Actual`).
- **SDMX 3.1**: the role attribute is **removed everywhere**. A `DataConstraint` only defines *allowed* content. Availability became a separate, dedicated object:
  - The [3.1 information model](https://sdmx.org/wp-content/uploads/SDMX_3-1-0_SECTION_2_FINAL.pdf) (§12.3, "AvailableDataConstraint") defines it as **not maintainable** ("generated dynamically based on the query", it "should not be (semantically) versioned") — it inherits from *Annotable* — and it "contains only 1 CubeRegion". This is the successor of the SDMX 2.1 `actualconstraint` REST resource / `type="Actual"` ContentConstraint.
  - [SDMX-ML 3.1](https://github.com/sdmx-twg/sdmx-ml/blob/v3.1.0/schemas/SDMXStructureConstraint.xsd) adds `AvailabilityConstraintType`: a required `ConstraintAttachment` referencing **exactly one** data-related artefact (`DataStructure` | `Dataflow` | `ProvisionAgreement`), **exactly one** `CubeRegion`, and optional `seriesCount`/`obsCount` attributes. Structure messages get a new `Structures/AvailabilityConstraints` container.
  - [SDMX-JSON 2.1](https://github.com/sdmx-twg/sdmx-json/blob/v2.1.0/structure-message/tools/schemas/sdmx-json-structure-schema.json) mirrors this: a `data.availabilityConstraints` collection with the same shape, and `role` removed from `dataConstraints`.

The bundled `sdmxschemas` package already ships the 3.1 XML and JSON schemas, so validation works out of the box.

### 1. Model: new `AvailabilityConstraint`; new time-range objects; `DataConstraint` untouched

```python
class TimePeriodBoundary(Struct, frozen=True, omit_defaults=True):
    """One end of a cube-region time range, with inclusivity."""
    period: str
    is_inclusive: bool = True

class CubeTimeRange(Struct, frozen=True, omit_defaults=True):
    """SDMX TimeRange: a period before/after, or a start/end range."""
    before_period: Optional[TimePeriodBoundary] = None
    after_period: Optional[TimePeriodBoundary] = None
    start_period: Optional[TimePeriodBoundary] = None
    end_period: Optional[TimePeriodBoundary] = None

class CubeKeyValue(Struct, frozen=True, omit_defaults=True):
    id: str
    values: Sequence[CubeValue] = ()            # now optional…
    time_range: Optional[CubeTimeRange] = None  # …mutually exclusive with time_range
    valid_from: Optional[datetime] = None       # SDMX-ML 3.0/3.1 only
    valid_to: Optional[datetime] = None         # SDMX-ML 3.0/3.1 only

class AvailabilityConstraint(Struct, frozen=True, omit_defaults=True):
    """The data actually present for one data-related artefact (SDMX 3.1)."""
    constraint_attachment: ConstraintAttachment  # exactly one DSD / dataflow / PA ref
    cube_region: CubeRegion                      # exactly one, per 3.1
    series_count: Optional[int] = None
    obs_count: Optional[int] = None
```

- `TimePeriodBoundary`, `CubeTimeRange` and the `CubeKeyValue` changes (`values` now optional and defaulting to `()`; new `time_range`, `valid_from`, `valid_to`) are exactly the objects added in PR #637 and carry over **as-is**.
- **`DataConstraint` gains no `role`/`type` field** (and no `ConstraintRole` enum is introduced). It keeps its current meaning, which is exactly the 3.1 one: allowed content.
- `AvailabilityConstraint` follows the 3.1 shape: annotable, not maintainable (no agency/id/version/name), single attachment, single cube region, no key sets.
- All exported from `pysdmx.model`.

### 2. Reading / writing matrix

| Format | Allowed content ↔ `DataConstraint` | Availability ↔ `AvailabilityConstraint` |
| --- | --- | --- |
| SDMX-ML 2.1 | `ContentConstraint type="Allowed"` — the writer must **always emit** the attribute, since it defaults to `Actual` | `ContentConstraint type="Actual"` or `type` **absent** (schema default) |
| SDMX-ML 3.0 | `DataConstraint role="Allowed"` (attribute required; treat absent as Allowed when reading lax files) | `DataConstraint role="Actual"` |
| SDMX-ML 3.1 | `DataConstraint` (no attribute) | `AvailabilityConstraint` in the new `AvailabilityConstraints` container, incl. `seriesCount`/`obsCount` |
| SDMX-JSON 2.0 | `dataConstraints` entry with `role: "Allowed"` or no `role` | `dataConstraints` entry with `role: "Actual"` |
| SDMX-JSON 2.1 | `dataConstraints` entry (no `role`) | `availabilityConstraints` entry |

Documented lossy edges of the legacy mappings:

- Reading a 2.1/3.0 "Actual" constraint **drops its maintainable identity** (agency, id, version, names) — per the 3.1 IM these objects are dynamically generated and unversioned.
- Writing an `AvailabilityConstraint` to 2.1/3.0/JSON 2.0 requires a maintainable artefact, so the writer **synthesises the identification** deterministically from the attached artefact (agency/id/version of the referenced dataflow/DSD/PA, generated name).
- `series_count`/`obs_count` can only be written to SDMX-ML 3.1 / SDMX-JSON 2.1; they are dropped for older versions.
- A legacy "Actual" constraint carrying key sets or more than one cube region does not fit the 3.1 object; the reader raises `Invalid` with a clear message (none observed in real files so far — can be revisited if they surface).

The `api/dc` availability flow (SDMX-JSON 2.0 responses from the REST `availability` endpoint) must keep working, now backed by `AvailabilityConstraint`.

### 3. Validity / TimeRange — what each format supports

A cube-region `KeyValue` is a schema **choice**: either one-or-more `Value`s **or** a single `TimeRange` (typically used for the time dimension):

```xml
<common:KeyValue id="TIME_PERIOD">
  <common:TimeRange>
    <common:StartPeriod isInclusive="true">1989-01-01T00:00:00</common:StartPeriod>
    <common:EndPeriod isInclusive="true">2024-12-31T00:00:00</common:EndPeriod>
  </common:TimeRange>
</common:KeyValue>
```

- **SDMX-ML 2.1**: a cube value (`SimpleValueType`) carries only `cascadeValues` — **no** `validFrom`/`validTo`. Time coverage is expressed **only** via `TimeRange`. Data keys have no validity either.
- **SDMX-ML 3.0/3.1**: `validFrom`/`validTo` are **optional on the cube-region `KeyValue`** (`MemberSelectionType`), **prohibited on data keys** (`DataKeyValueType`), and there is still **no per-`Value` validity**.
- Per-`Value` validity (`CubeValue.valid_from/valid_to`) and `DataKey` validity remain SDMX-JSON-only; the SDMX-ML writers cannot represent them (documented in code).
- Read/write `validFrom`/`validTo` at the cube-`KeyValue` level for SDMX-ML 3.0/3.1, and round-trip `TimeRange` in all formats.
- This reader-writer work is already implemented in PR #637 — SDMX-ML 2.1/3.0/3.1 `TimeRange` round-trip, KeyValue-level validity for 3.0/3.1, and SDMX-JSON `timeRange` + KeyValue validity — and carries over **as-is**.

### 4. Tests + samples

PR #637's test samples carry over **as-is** — 2.1 reader samples (`constraint_actual.xml`, `constraint_allowed.xml`, `constraint_no_type.xml`, `constraint_time_range.xml`), 3.0 reader samples (`constraint_role_absent.xml`, `constraint_time_range.xml`), the updated 2.1 writer samples (explicit `type` attribute) and the SDMX-JSON ser/deser tests — with the role-based assertions re-pointed at the two-class mapping. Required coverage:

- 2.1 reader: `type="Actual"`, `type="Allowed"`, `type` absent (→ availability, per the schema default).
- 3.0 reader: `role="Actual"`, `role="Allowed"`, `role` absent.
- 3.1 reader + writer: native `AvailabilityConstraint` element (attachment, cube region, counts), schema-valid output.
- SDMX-JSON: 2.0 `role` mapping in both directions; 2.1 `availabilityConstraints` read/write.
- `CubeRegion` with `TimeRange` and KeyValue-level validity (3.0/3.1), round-tripping with `validate=True` for 2.1/3.0/3.1.
- The OECD file above parses end-to-end with nothing dropped (20 structures; the availability constraint keeps its `TimeRange` 1989→2024).

## Relation to PR #637

PR #637 (draft) implemented the previous, role-based scope. The `TimePeriodBoundary`/`CubeTimeRange` objects, the `CubeKeyValue` changes, the TimeRange/validity reader-writer work and the test samples carry over as-is. The `ConstraintRole` enum and the `role` field on `DataConstraint` are **dropped** and replaced by the `AvailabilityConstraint` mapping above (including the 3.1 writer, which no longer needs to silently discard the role).

javihern98:
Closes #635.

Aligns the constraint model with **SDMX 3.1**: data availability is a separate `AvailabilityConstraint` object, and `DataConstraint` carries **no role/type property**. Retains the cube-region `TimeRange`/validity support from the earlier iteration of this PR (the `ConstraintRole` enum + `role` field from that iteration are removed — never released).

## What changed

- **Model** (`model/constraint.py`): new `AvailabilityConstraint` — annotable, **not maintainable** (per the 3.1 IM: generated dynamically, unversioned), attached to exactly one DSD/dataflow/provision agreement, exactly one cube region, optional `series_count`/`obs_count`. New `TimePeriodBoundary`/`CubeTimeRange`; `CubeKeyValue` gains `time_range`, `valid_from`/`valid_to`, and `values` becomes optional (`values` and `time_range` are now mutually exclusive — `Invalid` if both are set). `DataConstraint` = allowed values only.
- **SDMX-ML 2.1/3.0**: readers map the legacy markers to the two classes — 2.1 `type` (absent → `Actual`, the schema default), 3.0 `role` (absent → `Allowed`). Writers always emit `type="Allowed"` (2.1) / `role="Allowed"` (3.0) for a `DataConstraint`; an `AvailabilityConstraint` is written as a legacy `Actual` constraint with identification synthesised from the attached artefact (its agency/id/version, name `Availability for {id}`).
- **SDMX-ML 3.1**: native `AvailabilityConstraints` container, read and write, including `seriesCount`/`obsCount`; `DataConstraint` carries no marker. Round-trips are schema-valid (`validate=True`).
- **SDMX-JSON**: 2.0 `role` mapped in both directions (availability written as a legacy `role="Actual"` entry with synthesised identity); 2.1 writes native `availabilityConstraints` and omits `role` from `dataConstraints` (removed from the 2.1 schema).
- **`write_sdmx`** facade and all structure-writer signatures accept `AvailabilityConstraint`; `StructureMessage` gains `get_availability_constraints()`.

## Availability counts and the FMR `sdmx_metrics` convention

SDMX 3.1 carries `seriesCount`/`obsCount` as dedicated fields, but the legacy representations (SDMX-ML 2.1/3.0, SDMX-JSON 2.0) write availability as a plain `Actual` constraint, which has no such fields. Instead of inventing a pysdmx-only encoding, the legacy writers append one annotation per count in the shape the FMR itself emits on these constraints: `type="sdmx_metrics"`, `id` = `series_count`/`obs_count`, and the count as a string in `title` — e.g. `{"id": "obs_count", "type": "sdmx_metrics", "title": "42"}`. This is the same convention pysdmx already parses to fill `DataflowInfo.series_count`/`obs_count` (`__parse_annotation_metrics` in the SDMX-JSON dataflow reader, used for the FMR dataflow-details query), so FMR-produced documents are understood out of the box.

On read of a legacy `Actual` constraint, the XML and JSON readers lift such annotations into the `series_count`/`obs_count` model fields and drop them from `annotations`, so they are not written back out as duplicates. Only a `title` that is a plain non-negative integer (`str.isdecimal()`) is lifted; anything else stays a regular annotation. The native formats never synthesise or lift these annotations — counts live in their dedicated fields and annotations pass through untouched.

## Documented edges

- Reading a legacy 2.1/3.0 `Actual` constraint drops its maintainable identity (3.1: dynamically generated, unversioned). Legacy `Actual` constraints with key sets, ≠1 cube region, or no attachment raise `Invalid`.
- `series_count`/`obs_count` round-trip in every format: natively in SDMX-ML 3.1 / SDMX-JSON 2.1, and via the `sdmx_metrics` annotations described above elsewhere.
- Availability annotations are serialised everywhere: natively in SDMX-ML 3.1 / SDMX-JSON 2.1, and on the synthesised legacy constraints in 2.1/3.0/JSON 2.0.
- Two availability constraints for the same artefact would share a synthetic short URN, so readers and writers reject that case with `Invalid` instead of silently keeping the last one.
- A document setting both `Value`s and a `TimeRange` on one cube key value raises `Invalid` on read (previously one side was silently dropped at write time).
- A `CubeTimeRange` must match one of the three SDMX shapes (before only, after only, or start and end) — anything else raises `Invalid`, including on read of malformed documents.

## Verification

Full suite: **5187 passed, 100% branch coverage**; `ruff format`, `ruff check`, `mypy` clean. The io/model tests also pass on a real Python 3.10 interpreter.

Real-world end-to-end (the file from #635):

```python
from pysdmx.model import AvailabilityConstraint

# https://sdmx.oecd.org/public/rest/dataflow/OECD.SDD.TPS/DSD_SDBSBD_ISIC4@DF_BD_API/1.0?references=all
msg = read_sdmx("oecd_structures.xml", validate=False)
acs = [s for s in msg.structures if isinstance(s, AvailabilityConstraint)]
assert len(msg.structures) == 20 and len(acs) == 1
tr = next(
    kv.time_range
    for kv in acs[0].cube_region.key_values
    if kv.id == "TIME_PERIOD"
)
print(acs[0].reference)
print(tr.start_period.period, "->", tr.end_period.period)
```

Output: the dataflow URN for `OECD.SDD.TPS:DF_BD_API(1.0)`, then `1989-01-01T00:00:00 -> 2024-12-31T00:00:00` — all 20 structures parse and the availability constraint keeps its `TIME_PERIOD` coverage.

## Follow-ups (out of scope here)

Reader-built model objects carry **list**-typed sequence fields while hand-built ones use tuples, so they can be unequal to identically-valued hand-built objects, or unhashable. Normalizing this codebase-wide is #680 — covering the reader construction sites, the Python 3.10 empty-tuple interning constraint they surfaced, and the `Sequence[...] = []` model defaults (`RepresentationMap.maps` and friends — the same behaviour from the defaults side). The reader part was carried here briefly (commits `406f7405`/`ef876427`, reverted) and can be cherry-picked. The new availability-constraint JSON paths already follow the tuple convention.


src/pysdmx/model/constraint.py
```diff
@@ -61,9 +94,27 @@ class KeySet(Struct, frozen=True, omit_defaults=True):
...
+
+    def __repr__(self) -> str:
+        """Constraint role string representation."""
+        return f"{self.__class__.__name__}.{self._name_}"
+
+
 class DataConstraint(MaintainableArtefact, frozen=True, omit_defaults=True):
     """A data constraint, defining the allowed or available values."""
 
+    role: ConstraintRole = ConstraintRole.ALLOWED
```
javihern98, line 117: Hi @sosna leaving this for your review as I believe this approach is the more correct one (adding a role parameter to DataConstraint object to ensure we support 2.1 and 3.0 SDMX-ML roundtrip). This is not written into SDMX 3.1 related formats (SDMX-JSON 2.1 and SDMX-ML 3.1). 

We always discussed the model should reflect SDMX 3.X, but this seems to be a breaking change between SDMX 3.0 and 3.1, so I found this to be the best solution.

javihern98:
@sosna @stratosn This is now ready for your review. I believe I documented properly all the small edge cases (which they are quite a few).

src/pysdmx/io/json/sdmxjson2/messages/constraint.py
```diff
@@ -171,20 +254,123 @@ def from_model(self, ks: KeySet) -> "JsonKeySet":
         )
 
 
+_METRIC_ANNOTATION_IDS = ("series_count", "obs_count")
+
+
+def _metric_annotations(
+    cons: AvailabilityConstraint,
+) -> "tuple[JsonAnnotation, ...]":
```
sosna: Why are the returned types between quotes?

src/pysdmx/io/json/sdmxjson2/messages/constraint.py
```diff
@@ -171,20 +254,123 @@ def from_model(self, ks: KeySet) -> "JsonKeySet":
...
+                title=str(cons.obs_count),
+                type="sdmx_metrics",
+            )
+        )
+    return tuple(metrics)
+
+
+def _lift_metric_annotations(
+    annotations: Sequence[JsonAnnotation],
+) -> "tuple[Optional[int], Optional[int], tuple[Annotation, ...]]":
```
sosna: Why are the returned types between quotes?

src/pysdmx/io/json/sdmxjson2/writer/v2_1/structure.py
```diff
@@ -4,20 +4,22 @@
...
 
 from pysdmx.io.json.sdmxjson2.writer._helper import write_structure_msg
+from pysdmx.model import AvailabilityConstraint
 from pysdmx.model.__base import MaintainableArtefact
 from pysdmx.model.message import Header
 
 
 def write(
-    structures: Sequence[MaintainableArtefact],
+    structures: Sequence[Union[MaintainableArtefact, AvailabilityConstraint]],
```
sosna, line 13: I don't understand this. `AvailabilityConstraint` is **not** a maintainable artefact. Why would we ever need to write it?

src/pysdmx/io/json/sdmxjson2/writer/_helper.py
```diff
@@ -67,16 +67,17 @@ def write_metadata_msg(
 
 
 def write_structure_msg(
-    structures: Sequence[MaintainableArtefact],
+    structures: Sequence[Union[MaintainableArtefact, AvailabilityConstraint]],
```
sosna, line 70: I don't understand this. AvailabilityConstraint is not a maintainable artefact. Why would we ever need to write it?

src/pysdmx/io/xml/sdmx21/writer/structure.py
```diff
@@ -14,20 +15,22 @@
...
     get_end_message,
 )
+from pysdmx.model import AvailabilityConstraint
 from pysdmx.model.__base import MaintainableArtefact
 from pysdmx.model.message import Header
 
 
 def write(
-    structures: Sequence[MaintainableArtefact],
+    structures: Sequence[Union[MaintainableArtefact, AvailabilityConstraint]],
```
sosna, line 24: Same comment as above about the need to write AvailabilityConstraint

src/pysdmx/io/xml/sdmx30/writer/structure.py
```diff
@@ -14,20 +15,22 @@
...
     get_end_message,
 )
+from pysdmx.model import AvailabilityConstraint
 from pysdmx.model.__base import MaintainableArtefact
 from pysdmx.model.message import Header
 
 
 def write(
-    structures: Sequence[MaintainableArtefact],
+    structures: Sequence[Union[MaintainableArtefact, AvailabilityConstraint]],
```
sosna, line 24: Same comment as above about the need to write AvailabilityConstraint

src/pysdmx/io/xml/sdmx31/writer/structure.py
```diff
@@ -14,20 +15,22 @@
...
     get_end_message,
 )
+from pysdmx.model import AvailabilityConstraint
 from pysdmx.model.__base import MaintainableArtefact
 from pysdmx.model.message import Header
 
 
 def write(
-    structures: Sequence[MaintainableArtefact],
+    structures: Sequence[Union[MaintainableArtefact, AvailabilityConstraint]],
```
sosna, line 24: Same comment as above about the need to write AvailabilityConstraint

src/pysdmx/io/xml/__structure_aux_writer.py
sosna, line 1: Same comment as above about the need to write AvailabilityConstraint

src/pysdmx/io/xml/__write_aux.py
sosna, line 1: Same comment as above about the need to write AvailabilityConstraint

src/pysdmx/io/writer.py
sosna, line 1: Same comment as above about the need to write AvailabilityConstraint

src/pysdmx/util/__init__.py
```diff
@@ -34,7 +34,7 @@ def parse_urn(urn: str) -> Union[ItemReference, Reference]:
                     return parse_short_item_urn(urn)
                 except Invalid:
                     raise Invalid(
-                        NF, "{urn} does not match any known pattern"
+                        NF, f"{urn} does not match any known pattern"
```
sosna, line 37: Good catch, thanks :).

sosna (commented):
Hi @javihern98. Here is my first batch of comments. When we'll have agreed on the need to write (or not) availability constraints, I will continue the review of the tests.

src/pysdmx/io/json/sdmxjson2/messages/constraint.py
javihern98: I checked again and removed the quotes, I remember it was something related to mypy, but seems all checks are passing

src/pysdmx/io/json/sdmxjson2/messages/constraint.py
javihern98: I checked again and removed the quotes, I remember it was something related to mypy, but seems all checks are passing

src/pysdmx/util/__init__.py
javihern98: :blush:

javihern98:
Discussing internally if we need to write the AvailabilityConstraint objects, given they are not maintainable

javihern98:
After discussion, we agreed to keep the possibility to write AvailabilityConstraints, but seems we need to make a small change in the FMR maintenance put_structures method, as FMR does not support them:

image

src/pysdmx/io/json/sdmxjson2/writer/v2_1/structure.py
javihern98: After discussion we agreed to write it as well and keep it in the StructureMessage. See https://github.com/bis-med-it/pysdmx/pull/637#issuecomment-5599709532

src/pysdmx/io/json/sdmxjson2/writer/_helper.py
javihern98: After discussion we agreed to write it as well and keep it in the StructureMessage. See https://github.com/bis-med-it/pysdmx/pull/637#issuecomment-5599709532

src/pysdmx/io/xml/sdmx21/writer/structure.py
javihern98: After discussion we agreed to write it as well and keep it in the StructureMessage. See https://github.com/bis-med-it/pysdmx/pull/637#issuecomment-5599709532

javihern98:
@sosna Please review this again for merge, I solved all conflicts

src/pysdmx/api/fmr/maintenance.py
```diff
@@ -158,7 +159,9 @@ def __post(
 
     def put_structures(
         self,
-        artefacts: Sequence[MaintainableArtefact],
+        artefacts: Sequence[
+            Union[MaintainableArtefact, AvailabilityConstraint]
```
sosna: I think this is wrong. On the contrary, we should make it explicit that we only accept maintainable artefacts

src/pysdmx/api/fmr/maintenance.py
```diff
@@ -169,14 +172,32 @@ def put_structures(
...
         """
         if not header:
             header = Header()
-        message = StructureMessage(header=header, structures=artefacts)
+        structures = [
+            a for a in artefacts if not isinstance(a, AvailabilityConstraint)
+        ]
+        if len(structures) != len(artefacts):
+            warnings.warn(
+                "Availability constraints are generated by the FMR and cannot "
```
sosna: Maybe we could rephrase this, as follows: "Availability constraints are not maintainable artefacts and, therefore, they are not meant to be stored in the FMR"?

sosna (changes requested):

src/pysdmx/api/fmr/maintenance.py
javihern98: Change is due to the internal handling we do of AvailabilityConstraints and the typing of the StructureMessage.structures property (as stated in the docstring), but I guess we can make it explicit by limiting the typing to just MaintainableArtefact

src/pysdmx/api/fmr/maintenance.py
javihern98: Ok, I will edit it

src/pysdmx/api/fmr/maintenance.py
javihern98: Changed to the explicit MaintainableArtefact typing, with the guard for AvailabilityConstraint still present to raise a Warning

sosna (commented):
Thanks a lot, @javihern98, this looks very good. 

I may have just identified one minor issue and I tried reproducing it with a test. Do you agree with it? If so, after the issue has been fixed, I can approve the PR.

Very nice and clean work, as always :)

javihern98:
Hi @sosna I managed to solve it, fortunately it was just a one-line change. Two comments to review:

1. In SDMX-JSON 2.1 the availability constraints live in `availabilityConstraints` with native `seriesCount`/`obsCount`, but `_extract_metrics` only looks at `dataConstraints`, so a 2.1 round trip still leaves the dataflow counts empty.
2. `_extract_metrics` picks the first constraint attached to the dataflow regardless of role, so an Allowed constraint listed before the Actual one would hide the metrics.

They could be addressed in a separate issue if you find it necessary

sosna (approved):
Perfect, thanks a lot, @javihern98

sosna:
> * In SDMX-JSON 2.1 the availability constraints live in `availabilityConstraints` with native `seriesCount`/`obsCount`, but `_extract_metrics` only looks at `dataConstraints`, so a 2.1 round trip still leaves the dataflow counts empty.
> * `_extract_metrics` picks the first constraint attached to the dataflow regardless of role, so an Allowed constraint listed before the Actual one would hide the metrics.

Good catch, thanks! I have added an issue for these

Issue #639, opened 2026-07-13
SDMX-JSON: strict date-time validation + normalize SDMX datetimes to timezone-aware


javihern98:
## Summary

Follow-up to #638. That change wired the SDMX-JSON 2.1 schema into the reader and enabled `jsonschema`'s format checker, but deliberately with the **stdlib `date` checker only** (no `rfc3339-validator`). It did not enforce strict RFC 3339 `date-time`, because doing so rejects timezone-less datetimes that pysdmx itself emits.

## Problem

- pysdmx reads tz-less SDMX datetimes (e.g. `coreRepresentation.format.startTime = "2000-01-01T00:00:00"`) into **naive** `datetime` objects (`Facets.start_time` / `end_time`) and writes them back tz-less.
- `StandardTimePeriodType` and `meta.prepared` in the SDMX-JSON schemas require a timezone under strict RFC 3339 `format: date-time`.
- So enabling `rfc3339-validator` breaks round-trip validation of pysdmx's **own** writer output (3 SDMX-JSON 2.0 write tests fail, plus the 2.1 writer output).
- Consequences of the current date-only checker: a pure-date `meta.prepared` (e.g. `2024-01-01`) is falsely rejected (matches both `date` and the unchecked `date-time`, so `oneOf` fails), and a malformed `date-time` is not caught by the schema.

## Scope

1. Add `rfc3339-validator` to the `json` / `all` extras to strictly enforce `date-time`.
2. Normalize SDMX datetimes to timezone-aware UTC consistently, so writer output is RFC 3339 compliant **and** round-trip model equality still holds. This spans:
   - sdmxjson reader (`get_facets`) and writer (`JsonTextFormat.from_model`)
   - fusion reader
   - the XML reader (for cross-format consistency)
   - `Facets.start_time` / `end_time` become timezone-aware — a public model behavior change
   - update the FMR test assertions (`tests/api/fmr/concept_checks.py`)
3. Decide and document the convention for tz-less inputs (assume UTC).

## Notes

- This is a public behavior change (facet datetimes gain `tzinfo`), so it warrants its own design and PR rather than being bundled into the schema-wiring change.

javihern98:
This is being currently discussed internally...

javihern98:
After discussing it, we will follow the current scope of this issue, with two minor tweaks:

- The `rfc3339-validator` needs to be included on the json extra to ensure we include timezone-aware date times
[...]

PR #640, opened 2026-07-13
Validate SDMX-JSON 2.1 against the 2.1 JSON Schema

### LINKED ISSUE #638: SDMX-JSON: validate 2.1 messages against the 2.1 JSON Schema
## Summary

SDMX-JSON **2.1** structure and reference-metadata messages are currently validated against the **2.0** JSON Schema. Our pinned `sdmxschemas>=1.0.0` ships JSON schemas only for `sdmx10`/`sdmx20`; there is no 2.1 schema. The reader's schema picker (`_schema_for` in `io/json/sdmxjson2/reader/doc_validation.py`) selects by filename, and since the 2.0/2.1 schema files share the same name, 2.1 messages silently fall back to the 2.0 schema.

`sdmxschemas` **1.1.0** now ships `json/sdmx21/…` and exposes `SDMX_JSON_21_{DATA,STRUCTURE,METADATA}_PATH`. Bump the floor and wire the reader to pick the schema matching each message's version.

## Changes

1. Bump `sdmxschemas>=1.1.0` in the `json`, `xml`, and `all` extras; refresh `poetry.lock`.
2. Import the `SDMX_JSON_21_*` paths and make `_schema_for` version-aware (2.1 vs 2.0, derived from `meta.schema`).
3. Enable `jsonschema`'s format checker on the validator. The 2.1 schema types `meta.prepared` as `oneOf: [date-time, date]`; without a format checker a datetime string matches both branches and `oneOf` fails. The stdlib `date` checker (no extra dependency) disambiguates it.
4. Tests: version-aware schema selection + a real 2.1 message validated against the 2.1 schema. Keep 100% coverage.

## Notes

- Selection stays keyed on `meta.schema` (contains `.../2.0.0/…`, `.../2.1/…`, or `.../2.1.0/…`), consistent with the existing detector in `input_processor.py`.
- **Strict `date-time` (RFC 3339) enforcement is deliberately out of scope here** and tracked in #639: it would require `rfc3339-validator`, but that rejects the timezone-less datetimes pysdmx itself emits (e.g. `coreRepresentation.startTime`), breaking writer round-trips. Fixing that means normalizing SDMX datetimes to tz-aware across readers/writer and is a public model behavior change.
- SDMX-JSON *data* messages remain out of scope (no data reader exists).

javihern98:
## Summary

Validate SDMX-JSON **2.1** messages against the SDMX-JSON 2.1 JSON Schema. Previously 2.1 structure / reference-metadata messages were silently validated against the **2.0** schema — `_schema_for` matched schemas by file name, and the 2.0/2.1 schema files share the same name.

## Changes

- Bump `sdmxschemas` to `>=1.1.0` (ships `json/sdmx21/…` and `SDMX_JSON_21_*` paths) in the `json`, `xml`, and `all` extras; refresh `poetry.lock`.
- Make `_schema_for` version-aware (2.1 vs 2.0, derived from `meta.schema`).
- Enable `jsonschema`'s format checker so the 2.1 `meta.prepared` `oneOf(date-time, date)` resolves correctly via the stdlib `date` checker (no new dependency).
- Add tests for version-aware schema selection and a real 2.1 message validated against the 2.1 schema.

Strict RFC 3339 `date-time` enforcement is intentionally deferred to #639: it requires `rfc3339-validator`, which rejects the timezone-less datetimes pysdmx itself emits, so it needs a broader (public) datetime-normalization change.

Closes #638.


sosna (approved):
Nice, thanks :)

PR #645, opened 2026-07-14
Preserve full datetimes when reading SDMX-CSV

### LINKED ISSUE #644: SDMX-CSV reader truncates full datetimes (hh:mm:ss) at the first colon
## Description

When reading an SDMX-CSV file, a full datetime value such as
`2025-12-19T14:30:00Z` (e.g. in an `EMBARGO_TIME` attribute) is truncated to
`2025-12-19T14` — everything from the first `:` onward is lost.

## Steps to reproduce

```python
from pysdmx.io.csv.sdmx10.reader import read

csv = (
    "DATAFLOW,FREQ,TIME_PERIOD,OBS_VALUE,EMBARGO_TIME\n"
    "MD:DF_1(1.0),M,1999-01,1,2025-12-19T14:30:00Z\n"
)
print(read(csv)[0].data["EMBARGO_TIME"].iloc[0])
```

## Expected

```
2025-12-19T14:30:00Z
```

## Actual

```
2025-12-19T14
```

## Root cause

The SDMX-CSV readers strip `id: name` labels (produced by `labels=both`
requests) by splitting each value on a bare `:`. The actual SDMX-CSV label
separator is `": "` (colon + space), and this splitting ran unconditionally,
so any legitimate `:` in a value — such as the time part of a datetime — was
discarded.

Affects the SDMX-CSV 1.0, 2.0 and 2.1 readers.

javihern98:
Closes #644

## Problem

Reading SDMX-CSV truncated full datetimes: `2025-12-19T14:30:00Z` →
`2025-12-19T14`. The readers stripped `id: name` labels by splitting values on
a bare `:`, discarding the time part of any datetime — and the split ran even
when no labels were present.

## Fix

Use the real SDMX-CSV label separator `": "` (colon + space) consistently in
readers and writers. Codes never contain spaces, so splitting on `": "` removes
a label while leaving colon-containing values (datetimes) intact, even with
`labels=both`.

- Readers 1.0/2.0/2.1: strip labels only from labelled columns (`": "` in the
  header); the structure-id name is stripped once, via `count(": ")`.
- Writers: emit `": "` for labelled column headers, values and the
  structure-id name.
- Also fixes label-stripping being skipped on the first data column for
  SDMX-CSV 2.x files without an `ACTION` column.

See `labels=both` parameter in [SDMX-CSV field guide](https://github.com/sdmx-twg/sdmx-csv/blob/master/data-message/docs/sdmx-csv-field-guide.md#optional-parameters)

## Tests

Added datetime reader tests and write→read roundtrip tests (labels
`None`/`id`/`both`); updated `labels=both` fixtures to the `": "` format. Full
suite green at 100% branch coverage; `ruff` and `mypy` clean.


sosna (approved):
Good catch, thanks, @javihern98

PR #650, opened 2026-07-23
Align SDMX-CSV labels=name and keys columns with the field guide

### LINKED ISSUE #649: SDMX-CSV labels=name support does not follow the field guide
The SDMX-CSV field guide (data message, 2.0.0/2.1) defines `labels=name` as: *the id/value and the name of the artefacts are displayed in separate columns, the ID/value column always directly preceding its related localised name column*, plus a `STRUCTURE_NAME` column right after `STRUCTURE_ID`. Example from the field guide:

```
STRUCTURE,STRUCTURE_ID,STRUCTURE_NAME,ACTION,DIM_1,Dimension 1,DIM_2,Dimension 2,OBS_VALUE,Observation value
dataflow,ESTAT:NA_MAIN(1.6.0),National Accounts Main Aggregates,I,A,Value A,B,Value B,12.4,
```

pysdmx currently deviates from the specification:

- **Writer** (`labels="name"`): renames each component column to the concept name (single column, ids lost) instead of writing id+name column pairs.
- **Reader**: only drops `STRUCTURE_NAME`; the name columns of a spec-compliant `labels=name` file are kept as extra data columns instead of being removed.
- **Keys columns**: the writer emits `SERIES_KEYS`/`OBS_KEYS` (plural) while the spec — and the writer's own docstring — says `SERIES_KEY`/`OBS_KEY`.

Proposed changes:

- Reader: detect `labels=name` via `STRUCTURE_NAME` and drop the paired name columns, keeping only the ids; raise `Invalid` on a malformed (odd) component column count.
- New `drop_labels` utility to strip labels (`labels=both` and `labels=name` formats) from plain pandas DataFrames, reused by the CSV readers.
- Writer: emit spec-compliant id+name column pairs for `labels="name"`; raise `Invalid` when duplicate column names would be produced.
- Writer and reader: use the spec-compliant `SERIES_KEY`/`OBS_KEY` column names (the plural forms produced by previous versions were a writer bug and are not supported).

javihern98:
Closes #649

Reviewed the `labels=name` support against the [SDMX-CSV field guide](https://github.com/sdmx-twg/sdmx-csv/blob/master/data-message/docs/sdmx-csv-field-guide.md): the spec requires separate id+name column pairs (id column directly followed by its localised-name column) and a `STRUCTURE_NAME` column after `STRUCTURE_ID`. pysdmx instead renamed id columns to names on write and kept the name columns as data on read.

### Changes

- **Reader**: detects `labels=name` via `STRUCTURE_NAME` and drops the name columns positionally, keeping only the ids; raises `Invalid` when the id/name pairing is broken (odd component column count).
- **`drop_labels`** (new, `pysdmx.toolkit.pd`): strips `labels=both` or `labels=name` from a plain pandas DataFrame; the three CSV readers now share this single implementation.
- **Writer**: `labels="name"` inserts a localised-name column right after each component column; raises `Invalid` if duplicated column names would be produced (e.g. a concept name equal to a component id).
- **Keys columns**: writer and reader use the spec-compliant `SERIES_KEY`/`OBS_KEY` names (as the writer docstring already stated).
- Test DSD concept `OBS_VALUE` renamed to `Observation Value` (was a placeholder equal to the id).

### Behavior changes

- Files written by previous pysdmx versions with `labels="name"` (non-compliant single name-headed columns) no longer read back cleanly — that format was spec-non-compliant and lossy (ids were gone).
- `keys=` output columns renamed `SERIES_KEYS`/`OBS_KEYS` → `SERIES_KEY`/`OBS_KEY`; the old plural columns are not handled on read (they were a writer bug).
- `write(labels="name")` raises `Invalid` on non-unique concept names instead of silently producing a broken file.


sosna (approved):

sosna:
Looks good! Thanks, @javihern98!

PR #652, opened 2026-07-24
Prevent bugs on SDMX-ML reader on constraints with Actual/Allowed marker and time-range cube regions

### LINKED ISSUE #651: Read SDMX-ML DataConstraints with an Actual/Allowed marker and time-range cube regions
## Problem

The SDMX-ML structure reader cannot read two kinds of `DataConstraint` that real
`.Stat` / .NET registries return (e.g. a `references=all` query yields the
auto-created *actual content constraint* `CR_A_<dataflow>`):

1. **Actual/Allowed marker.** The distinction is carried by `role` (SDMX 3.0) or
   `type` (SDMX 2.1). The reader raised `NotImplementedError` when it saw
   `role="Actual"`, so any message containing an actual constraint failed to read.
2. **Time-range / numeric cube regions.** A cube-region `KeyValue` can express a
   `TimeRange` or numeric `Range` instead of enumerated `<Value>` children. The
   reader assumed `<Value>` was always present and crashed (`KeyError`) on those.

The simplified pysdmx model represents neither the Actual/Allowed distinction nor
range-valued key-values, so the fix is to read such constraints as plain
`DataConstraint`s and skip the range key-values.

## Fix

Both changes are in `src/pysdmx/io/xml/__structure_aux_reader.py`:

- `__format_constraint`: drop both `role` (SDMX 3.0) and `type` (SDMX 2.1) markers
  (`element.pop(..., None)`) so a constraint of either kind reads as a plain
  `DataConstraint` — replacing the previous `NotImplementedError` on `role="Actual"`.
- Cube-region `KeyValue` loop: `if VALUE not in kv: continue` — skip a key-value
  that carries a `TimeRange`/numeric `Range` rather than enumerated values.

## Tests

- New sample `tests/io/xml/sdmx21/reader/samples/constraint_actual_timerange.xml`
  (an actual constraint mixing an enumerated FREQ key-value with a TIME_PERIOD
  TimeRange).
- New `tests/io/xml/sdmx21/reader/test_reader.py::test_constraint_actual_type_and_time_range`.
- Update the existing `tests/io/xml/sdmx30/reader/test_reader.py` test that asserted
  a raise on `role="Actual"` so it instead asserts a plain `DataConstraint` is read
  (reuses the existing `constraint_actual.xml` sample).

albertohernandez1995:
## Summary

The SDMX-ML reader failed on two kinds of `DataConstraint` that real .Stat /
.NET registries return:

- constraints carrying an **Actual/Allowed marker** (`role` in SDMX 3.0,
  `type` in SDMX 2.1) — previously raised `NotImplementedError` on
  `role="Actual"`;
- **time-range / numeric cube regions** — a `KeyValue` expressing a `TimeRange`
  or numeric `Range` instead of enumerated `<Value>`s crashed with `KeyError`.

## Fix

`src/pysdmx/io/xml/__structure_aux_reader.py`:
- `__format_constraint` drops both `role` and `type` markers, so a constraint of
  either kind reads as a plain `DataConstraint`.
- The cube-region loop skips a `KeyValue` with no `<Value>` (a range the
  simplified model does not represent).

## Tests

- New sample `constraint_actual_timerange.xml` +
  `sdmx21/.../test_constraint_actual_type_and_time_range`.
- `sdmx30` test updated from asserting a raise to asserting a plain
  `DataConstraint`.

---
Closes #651


sosna (approved):
It looks good. Thanks, @albertohernandez1995

PR #665, opened 2026-08-07
Improve error reported in case of SDMX-REST version mismatch

### LINKED ISSUE #664: Improve error reporting in REST API queries
Specifying `asOf` when constructing a structure query for SDMX-REST 2.0.0 returns an odd error: “as_of not supported in 8”. 

This is because  `asOf` is not supported in SDMX-REST 2.0.0 (It was introduced in version 2.2.0). However, a more descriptive message would be helpful.

sosna:
Close #664


javihern98 (approved):
Looks good, thanks! 😊

PR #668, opened 2026-08-10
Accept case-insensitive values in the STRUCTURE column of SDMX-CSV

### LINKED ISSUE #667: Accept case-insensitive values in the STRUCTURE column of SDMX-CSV 2.0
## Description

The SDMX-CSV 2.0 reader only accepts exact lowercase values in the `STRUCTURE` column (`dataflow`, `datastructure`, `dataprovision`). Files using uppercase or mixed-case values (e.g. `DATAFLOW`, `Dataflow`) are rejected with an `Invalid` error.

The check is in `src/pysdmx/io/csv/__csv_aux_reader.py` (shared by the SDMX-CSV 2.0 and 2.1 readers, so both are affected):

```python
if structure_type == "DataStructure".lower():
    urn = f"DataStructure={structure_id}"
elif structure_type == "Dataflow".lower():
    urn = f"Dataflow={structure_id}"
elif structure_type == "dataprovision":
    urn = f"ProvisionAgreement={structure_id}"
else:
    raise Invalid(
        "Invalid value on STRUCTURE column",
        ...
    )
```

Note that `.lower()` is applied to the string literals rather than to the value read from the file, so the matching is effectively case-sensitive.

## Example

```python

csv_upper = """STRUCTURE,STRUCTURE_ID,FREQ,REF_AREA,OBS_VALUE
DATAFLOW,MD:TEST_DF(1.0),A,ES,12.3
"""

read_sdmx(csv_upper)
```

Output:

```
pysdmx.errors.Invalid: Invalid value on STRUCTURE column: Invalid SDMX-CSV file. Check the docs for the proper values on STRUCTURE column.
```

The same file with `dataflow` in lowercase is read correctly.

## Expected behaviour

While the SDMX-CSV specification writes these values in lowercase, the reader should be lenient with the casing used by other tools and accept the `STRUCTURE` column values case-insensitively (e.g. by comparing `structure_type.lower()` against the allowed values).

javihern98:
Closes #667

The SDMX-CSV 2.0 and 2.1 readers rejected files whose `STRUCTURE` column used non-lowercase values (e.g. `DATAFLOW`, `Dataflow`), because `.lower()` was applied to the string literals instead of the value read from the file.

The value read from the file is now lowercased before the comparison, so `STRUCTURE` values are matched case-insensitively, as suggested in the issue. Added tests for both readers covering the three structure types with non-lowercase casings.

Verified against real-world data — the OECD .Stat Suite writes `DATAFLOW` in uppercase. To reproduce:

```python
import httpx

url = (
    "https://sdmx.oecd.org/public/rest/data/"
    "OECD.SDD.STES,DSD_STES@DF_CLI,4.1/USA.M.LI...AA...H"
    "?lastNObservations=3"
)
headers = {"Accept": "application/vnd.sdmx.data+csv; version=2.0.0"}
sdmx_csv = httpx.get(url, headers=headers).text

datasets = read_sdmx(sdmx_csv).data
print(datasets[0].short_urn)  # Dataflow=OECD.SDD.STES:DSD_STES@DF_CLI(4.1)
```

Without this fix, the snippet raises `Invalid: Invalid value on STRUCTURE column`.


mikwespinoza-ux (dismissed):

sosna (approved):
Looks good! Thanks, @javihern98 :)

PR #669, opened 2026-08-11
Add an MCP server for SDMX data discovery and retrieval

markchweya:
## What this adds

An MCP ([Model Context Protocol](https://modelcontextprotocol.io)) server exposing
`pysdmx.api.dc.pd.PandasConnector` to AI assistants, under `pysdmx.toolkit.mcp`.

This relates to #561 ("AI Coding Assistant Instructions for data analysts"). That
issue asks for guidance aimed at analysts; this is one concrete form of it — the
behaviour described in `.agents/skills/data-discovery-and-retrieval/SKILL.md`
implemented as a server, so the rules are enforced by the tool rather than left for
an assistant to remember.

The same server is also published standalone as
[`sdmx-data-mcp`](https://pypi.org/project/sdmx-data-mcp/), which depends on
`pysdmx[data]` rather than living inside it:

```bash
pip install sdmx-data-mcp
claude mcp add sdmx -- sdmx-data-mcp
```

So this can live outside the library if you would prefer that, and I am happy to
close the PR on that basis. I am raising it here because enforcing the SKILL.md
rules in the tool seemed closer to what #561 is asking for than documentation
alone.

The distinguishing feature is that `get_data` returns observations. Existing SDMX
MCP servers navigate metadata and hand back a query URL, leaving the assistant with
a link rather than numbers.

## The four tools

Meant to be called in order:

| Tool | Purpose |
|---|---|
| `list_services` | Endpoints from `Endpoints`, plus any caller-supplied SDMX-REST v2 base URL |
| `search_dataflows` | One `dataflows()` call, then local term matching |
| `inspect_dataflow` | Components, availability-backed codes, size signals |
| `get_data` | Retrieves the observations |

## Design decisions worth reviewing

These are the points where I made a judgement call, and the ones most worth
pushing back on.

### The error hierarchy is preserved rather than flattened

`NotFound` and `Unavailable` imply opposite retry decisions, so each error maps to
a distinct kind with an explicit retriable flag. I marked `InternalError`
non-retriable, following the class docstring in `errors.py` ("clients should not
retry the query before investigating"), even though it maps to 5xx.

### `obs_count` is treated as optional

Against `BIS:WS_CBS_PUB(1.0)` it comes back `None` for both the full flow and
filtered subsets, while `series_count` is populated. Size checks therefore lead on
`series_count`; anything doing `if flow.obs_count > n` would raise `TypeError`
against the only endpoint `Endpoints` ships.

### Retrieval is capped

`L_MEASURE = 'S' AND L_REP_CTY = 'CH' AND FREQ = 'Q'` — three clauses, one country
— returns 478,965 rows. The default cap is 1,000 with a hard ceiling of 10,000, and
the response sets `truncated` and states that the rows are the first N in service
order rather than a sample.

### A code is located across every component carrying it

On consolidated banking, `CH` is available in `L_REP_CTY`, `L_CP_COUNTRY` and
`CBS_BANK_TYPE`, which shares the country codelist.
`inspect_dataflow(find_code=...)` reports all three with a role hint, so "in
Switzerland" cannot silently resolve to the wrong dimension.

### Time filters degrade rather than fail

`TIME_PERIOD` is pushed down first. If the service rejects it with `Invalid` or
`NotImplemented`, the clause is stripped, the narrower query retried, and the
cutoff applied in Pandas, with `filter_fallback` saying so. It deliberately does
not fall back on `NotFound`, `Unavailable` or `InternalError`, where dropping a
clause cannot help and would only obscure the real error. Pushdown does work
against BIS today; the fallback is covered by tests rather than by live behaviour.

### Availability is never reported as validity

Every inspection result carries a note that a code missing from availability may
still be valid in the full codelist.

## Packaging

- New `mcp` extra (`fastmcp>=3.0,<4.0`, gated to Python >= 3.10), added to `all`
- `__check_mcp_extra()` follows the existing `__extras_check.py` pattern
- New `mcp` pytest marker, registered in `pytest.ini` and auto-applied via
  `tests/conftest.py`
- Entry point: `python -m pysdmx.toolkit.mcp`, STDIO by default, `--transport http`
  for Streamable HTTP

## Verification

- `ruff format` and `ruff check` clean
- `mypy` clean across all 191 source files
- 5,256 tests pass; 100% statement and branch coverage repo-wide, with the new
  package at 100% on its own
- Sphinx docs build clean; API reference and how-to added to the toolkit toctrees

Most server tests inject a fake connector so every branch is reachable
deterministically. A separate end-to-end module drives the real `PandasConnector`
against `respx`-mocked responses, reusing the existing `bis_der.json` and
`data_v2.csv` samples, so drift in URL construction or SDMX-CSV parsing surfaces
there.

The API surface was re-read from v1.18.0 source and confirmed against the live BIS
service before the server was written.

### Behaviour against the live service

Driven from a client, `BIS:WS_CBPOL(1.0)` with
`FREQ = 'M' AND REF_AREA = 'CH' AND TIME_PERIOD >= '2020-01'` returns 79 rows with
`filter_fallback: null`, and the values reproduce the SNB rate path correctly
(−0.75 held to May 2022, peak of 1.75 from June 2023, 0.00 from June 2025).

Two behaviours showed up unprompted and are worth noting, since both are the
result of the design decisions above:

- A failing `inspect_dataflow` call returned `internal_error` with
  `retriable=false` and a `next_step` telling the caller not to repeat the
  identical request. The assistant changed approach rather than retrying.
- Asked for a daily KES series, the server returned `not_found` rather than an
  empty result, so the caller concluded the series does not exist instead of
  searching other date ranges for it.

---

*Developed with Claude Code. The design decisions above are mine and I am happy to
discuss any of them.*

### DIFF
--- a/docs/api/toolkit.rst
+++ b/docs/api/toolkit.rst
@@ -15,3 +15,4 @@ version of the SDMX standard.
    toolkit/vtl
    toolkit/pandas
    toolkit/sqlsrv
+   toolkit/mcp
--- a/docs/api/toolkit/mcp.rst
+++ b/docs/api/toolkit/mcp.rst
@@ -1,0 +1,30 @@
+MCP Server
+==========
+
+The MCP Toolkit exposes pysdmx data discovery and retrieval over the
+`Model Context Protocol <https://modelcontextprotocol.io>`_, so that AI
+assistants can find statistical datasets and retrieve the observations
+themselves.
+
+Requires the ``mcp`` and ``data`` extras::
+
+    pip install pysdmx[mcp,data]
+
+Run it over STDIO, the transport MCP clients launch::
+
+    python -m pysdmx.toolkit.mcp
+
+Or over Streamable HTTP::
+
+    python -m pysdmx.toolkit.mcp --transport http --port 8000
+
+Tools
+-----
+
+.. autofunction:: pysdmx.toolkit.mcp.list_services
+
+.. autofunction:: pysdmx.toolkit.mcp.search_dataflows
+
+.. autofunction:: pysdmx.toolkit.mcp.inspect_dataflow
+
+.. autofunction:: pysdmx.toolkit.mcp.get_data
--- a/docs/howto/toolkit/general.rst
+++ b/docs/howto/toolkit/general.rst
@@ -15,3 +15,5 @@ The toolkit functions does not follow any particular version of the SDMX standar
 :ref:`sqlsrv_toolkit`
 
 :ref:`pandas_toolkit`
+
+:ref:`mcp_toolkit`
--- a/docs/howto/toolkit/mcp_toolkit.rst
+++ b/docs/howto/toolkit/mcp_toolkit.rst
@@ -1,0 +1,200 @@
+.. _mcp_toolkit:
+
+MCP Server
+==========
+
+The MCP Toolkit turns pysdmx into a
+`Model Context Protocol <https://modelcontextprotocol.io>`_ server, so an
+AI assistant can discover SDMX dataflows and **retrieve the observations**,
+rather than being handed a query URL to fetch by itself.
+
+It wraps :class:`pysdmx.api.dc.pd.PandasConnector` and exposes four tools.
+
+Installation
+------------
+
+.. code-block:: bash
+
+    pip install pysdmx[mcp,data]
+
+The ``mcp`` extra brings in FastMCP; ``data`` brings in Pandas, which the
+underlying connector needs.
+
+Running the server
+------------------
+
+Most MCP clients launch the server themselves over STDIO:
+
+.. code-block:: bash
+
+    python -m pysdmx.toolkit.mcp
+
+To share one instance over the network instead, use Streamable HTTP:
+
+.. code-block:: bash
+
+    python -m pysdmx.toolkit.mcp --transport http --host 127.0.0.1 --port 8000
+
+Client configuration
+--------------------
+
+**Claude Desktop** — add to ``claude_desktop_config.json``:
+
+.. code-block:: json
+
+    {
+      "mcpServers": {
+        "sdmx": {
+          "command": "python",
+          "args": ["-m", "pysdmx.toolkit.mcp"]
+        }
+      }
+    }
+
+**Claude Code** — from the command line:
+
+.. code-block:: bash
+
+    claude mcp add sdmx -- python -m pysdmx.toolkit.mcp
+
+**Cursor** — add to ``.cursor/mcp.json``:
+
+.. code-block:: json
+
+    {
+      "mcpServers": {
+        "sdmx": {
+          "command": "python",
+          "args": ["-m", "pysdmx.toolkit.mcp"]
+        }
+      }
+    }
+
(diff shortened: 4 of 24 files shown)


markchweya:
### Update: this is now published and working in practice

To avoid this PR being the only way to try the server, I have released the same
code as a standalone package that depends on `pysdmx[data]` from PyPI. It does
not fork or vendor pysdmx, and uses only the public API — `PandasConnector`,
`Endpoints` and the errors hierarchy.

- PyPI: https://pypi.org/project/sdmx-data-mcp/
- Source: https://github.com/markchweya/sdmx-data-mcp

```bash
pip install sdmx-data-mcp
claude mcp add sdmx -- sdmx-data-mcp
```

That is entirely independent of this PR. I would still rather the capability
lived in pysdmx itself, and I am happy to close the package down or point it at
`pysdmx.toolkit.mcp` if you take this in.

### What real use has confirmed

Driving it from an assistant against the live BIS service, the parts of the
design this PR argued for do the work they were meant to:

- Retrieval is genuinely complete when it says so. `FREQ = 'M' AND REF_AREA =
  'CH' AND TIME_PERIOD >= '2020-01'` on `WS_CBPOL` returns 79 of 79 rows with
  `truncated: false`, so aggregating is safe.
- `IN (...)` covers the absence of `OR` without the assistant improvising.
- Reporting `obs_count` as optional was necessary, not defensive: it is `null`
  on every BIS scope observed so far, and `series_count` carries the signal.
- The ambiguity check earns its place. On `WS_CBS_PUB`, `find_code="CH"` still
  returns three components — reporting country, counterparty country, and the
  bank-type dimension that reuses the country codelist.

Worked transcripts, all captured output rather than illustrations:
https://github.com/markchweya/sdmx-data-mcp/blob/main/docs/EXAMPLES.md

### One genuine service behaviour worth flagging

`inspect_dataflow` on `BIS:WS_XRU(1.0)` with `REF_AREA = 'KE' AND FREQ = 'D'` —
a daily series that does not exist — comes back as:

```
[internal_error] Unexpected message format - The payload could not be
deserialized. This likely indicates that the service did not respond with a
valid SDMX-JSON v2.0.0 response.
```

So an availability query for an empty combination returns a response that is
not valid SDMX-JSON, rather than an empty result or a clean 404. pysdmx maps
that to `InternalError`, which is correct and non-retriable, and the server
passes it through with that meaning intact. Flagging it in case it is
unexpected on the BIS side rather than intended.

### Also in this branch

The latest commit fixes a portability bug the CI matrix caught: `DataFrame.where(..., other=None)`
does not type-check against the pandas-stubs resolved on Python 3.10, though it
passes on 3.11 and later. Missing values are now normalised explicitly, which
also covers `pd.NA` — the previous `notna` mask never reached it once dtypes
were cast to `object`.

Still 100% statement and branch coverage, with ruff and mypy clean.

javihern98:
Hi @markchweya, unfortunately the MCP server capabilities is out of scope for pysdmx (at least for now), while we acknowledge you may use pysdmx as the dependency of it in a separate package. 

Therefore we will now close this PR, thanks for your contribution :blush:

PR #671, opened 2026-08-13
Add case_mode option for WHERE clauses in SQL Server toolkit

### LINKED ISSUE #670: Add `case_mode` option when building WHERE clauses in SQL toolkit
[Operators](https://github.com/bis-med-it/pysdmx/blob/ea09e4c4c39e4030521cd09ff54e17bf7d4e8598/src/pysdmx/api/dc/query/_model.py#L22) such as `LIKE` and `NOT_LIKE` are currently implemented by forcing case-insensitive matching with `UPPER(...)` in pysdmx [SQL Server toolkit](https://github.com/bis-med-it/pysdmx/blob/ea09e4c4c39e4030521cd09ff54e17bf7d4e8598/src/pysdmx/toolkit/sqlsrv/__init__.py#L429-L438).

This is suboptimal for two reasons:

- it silently enforces case-insensitive behavior, although some users may expect or require case-sensitive matching;
- it can be redundant when case-insensitive behavior is already provided by the database or column collation, and may introduce an unnecessary performance cost.

To address this, add a new `case_mode` option in [`get_where_clause`](https://github.com/bis-med-it/pysdmx/blob/ea09e4c4c39e4030521cd09ff54e17bf7d4e8598/src/pysdmx/toolkit/sqlsrv/__init__.py#L253) with the following values:

- `insensitive`: force case-insensitive matching
- `sensitive`: force case-sensitive matching
- `default`: use the database/column collation behavior

For backward compatibility, the default should be `insensitive`.

This makes the behavior explicit, avoids unnecessary normalization in environments where collation already provides the desired behavior, and gives users control over case sensitivity.

sosna:
Fix #670


javihern98 (approved):
Looks good, thanks! 😊

PR #674, opened 2026-08-14
Lower minimum httpx version required

sosna:
This is necessary as some users use pysdmx in combination with [lseg-data](https://pypi.org/project/lseg-data/), which sets the following boundaries for httpx: `httpx>=0.18,<0.28`.


javihern98 (approved):
Looks good, thanks! 😊

Issue #615, opened 2026-06-03
SDMX-ML writer corrupts single-quoted VTL identifiers in Transformation expressions


javihern98:
## Summary

When writing a VTL `TransformationScheme` to **SDMX-ML**, single-quoted VTL identifiers inside the `<str:Expression>` body are corrupted into double-quoted string literals. This produces invalid VTL that FMR (and any VTL parser) rejects with *"expecting IDENTIFIER"*.

## Reproduction

A `check_datapoint` result whose expression escapes the reserved VTL keyword `errorlevel` (which **must** be single-quoted to be used as an identifier):

```
check_datapoint(ds, dp_ruleset) [calc errorcode := errorcode, 'errorlevel' := errorlevel]
```

When written to SDMX-ML, the `<str:Expression>` body becomes:

```
check_datapoint(ds, dp_ruleset) [calc errorcode := errorcode, "errorlevel" := errorlevel]
```

`'errorlevel'` (a single-quoted **identifier escape**) silently became `"errorlevel"` (a **string literal**), so the output is no longer valid VTL.

## Expected vs. actual

| | `errorlevel` rendered as |
| --- | --- |
| **Expected** (and what the model holds) | `'errorlevel'` |
| SDMX-ML 2.1 / 3.0 / 3.1 | `"errorlevel"` ❌ |
| SDMX-JSON 2.0 / 2.1 | `'errorlevel'` ✅ (unaffected) |

Only the SDMX-ML writers are affected; the model and the SDMX-JSON writers are correct.

## Root cause

In `_write_vtl` (`src/pysdmx/io/xml/__structure_aux_writer.py`, line 1862), the writer builds XML attributes via Python `repr` (`!r`, which yields single quotes) and then runs a **blanket** `outfile.replace("'", '"')` over the *entire* serialized chunk to convert those into XML double quotes. That blanket replace also rewrites the single quotes inside the element **text bodies** — the VTL code in `<Expression>`, `<RulesetDefinition>` and `<OperatorDefinition>`. Those bodies are emitted via `__escape_xml` (`src/pysdmx/io/xml/__write_aux.py`), which escapes `& < >` and standalone `"` (added in #365 for the identical class of bug) but **not** `'`. The single quotes therefore survive into the blanket replace and get corrupted.

`_write_vtl` is shared by the SDMX-ML 2.1, 3.0 and 3.1 writers, so all three versions are affected.

## Proposed fix

[...]

PR #679, opened 2026-08-27
Fix apostrophe corruption in all but the last annotation when writing SDMX-ML

### LINKED ISSUE #678: SDMX-ML writer corrupts apostrophes in all but the last annotation on an artefact
Writing a structure message replaces every apostrophe with a double quote in the text of any annotation that isn't the last one on an artefact. The document stays schema-valid, so nothing raises and nothing warns. The text is just quietly wrong.

Reproduced on 1.19.0:

```python
from pysdmx.model import Annotation, Code, Codelist

def emit(*annotations):
    cl = Codelist(id="CL_REPRO", agency="TEST", version="1.0", name="Repro",
                  annotations=list(annotations), items=[Code(id="X", name="X")])
    return write_sdmx([cl], sdmx_format=Format.STRUCTURE_SDMX_ML_2_1,
                      output_path="", prettyprint=True)

one = emit(Annotation(id="A1", title="Ireland's GDP"))
two = emit(Annotation(id="A1", title="Ireland's GDP"),
           Annotation(id="A2", title="second"))

print("Ireland's GDP" in one)   # True
print("Ireland's GDP" in two)   # False
print('Ireland"s GDP' in two)   # True
```

With one annotation the apostrophe survives. Add a second and the first one's title becomes `Ireland"s GDP`. AnnotationText and AnnotationURL go the same way.

The cause is in `__write_annotable`, in `pysdmx/io/xml/__structure_aux_writer.py`:

```python
outfile = f"{child1}<{ABBR_COM}:Annotations>"
for annotation in annotable.annotations:
    ...
    outfile += f"{child2}<{ABBR_COM}:Annotation id={annotation.id!r}>"
    outfile = outfile.replace("'", '"')
```

`{annotation.id!r}` quotes with apostrophes, so the replace is there to turn `id='X'` into `id="X"`. But it runs over the entire accumulated buffer on every iteration rather than the fragment just appended, so it also rewrites everything written before it — the title, text and URL of every earlier annotation. Only the final annotation on an artefact is ever safe.

This is the same cause as #615, which fixed it for VTL code bodies by escaping single quotes to `&apos;` there. That issue reasoned the blanket replace shouldn't be allowed to touch legitimate apostrophes in Names and Descriptions, and that reasoning holds — I checked, and Names and Descriptions are unaffected. Annotations just weren't part of that fix.

Building the attribute directly sidesteps it, and then the replace has nothing left to do:

```python
outfile += f'{child2}<{ABBR_COM}:Annotation id="{annotation.id}">'
```

Tested on 1.18.0 and 1.19.0, Python 3.14.

One thing I haven't chased: the same `!r`-plus-blanket-replace pattern appears at several other points in that file. I don't know whether any of those can be reached with text containing an apostrophe.

albertohernandez1995:
## Summary

Fixes #678 — the SDMX-ML structure writer silently corrupted apostrophes in every annotation on an artefact except the last one.

Root cause is as diagnosed in the issue: `__write_annotable` built the `id` attribute with `{annotation.id!r}` and then applied a blanket `outfile.replace("'", '"')`. The replace ran **inside** the per-annotation loop and over the **whole accumulated buffer**, so each newly appended annotation re-quoted every apostrophe already written. Only the final annotation escaped unscathed, and the document stayed schema-valid, so the corruption was silent.

## Changes included

**`pysdmx.io.xml.__structure_aux_writer`** — two fixes in `__write_annotable` and its element map:

- **`__write_annotable`** — the `Annotation id` attribute is now quoted explicitly and escaped via `__escape_xml`; the in-loop blanket replace is deleted. Matches the existing explicit-quoting convention already used for `DatePatternMap` attributes in the same module.
  ```diff
  -            outfile += f"{child2}<{ABBR_COM}:Annotation id={annotation.id!r}>"
  -        outfile = outfile.replace("'", '"')
  +            annotation_id = __escape_xml(annotation.id)
  +            outfile += f'{child2}<{ABBR_COM}:Annotation id="{annotation_id}">'
  ```
- `url` and `text` swapped so the emitted sequence is `AnnotationTitle, AnnotationType, AnnotationURL, AnnotationText`, per the `AnnotationType` complex type in `SDMXCommon.xsd`.

Before / after, for a codelist with three annotations:

```xml
<!-- before -->
<com:AnnotationTitle>It"s the first title</com:AnnotationTitle>
<com:AnnotationTitle>It"s the middle title</com:AnnotationTitle>
<com:AnnotationTitle>It's the last title</com:AnnotationTitle>

<!-- after -->
<com:AnnotationTitle>It's the first title</com:AnnotationTitle>
<com:AnnotationTitle>It's the middle title</com:AnnotationTitle>
<com:AnnotationTitle>It's the last title</com:AnnotationTitle>
```

### DIFF
--- a/src/pysdmx/io/xml/__structure_aux_writer.py
+++ b/src/pysdmx/io/xml/__structure_aux_writer.py
@@ -2318,8 +2318,9 @@ def _write_vtl(  # noqa: C901
     if isinstance(item_or_scheme, Item):
         label = ""
         nameable = __write_nameable(item_or_scheme, add_indent(indent))
-        attrib = nameable["Attributes"].replace("'", '"')
-        data = __export_intern_data(nameable)
+        attrib = nameable["Attributes"]
+        intern = __export_intern_data(nameable)
+        data = ""
 
         if isinstance(item_or_scheme, Ruleset):
             label = f"{ABBR_STR}:{RULE}"
@@ -2524,14 +2525,15 @@ def _write_vtl(  # noqa: C901
                 f"</{ABBR_STR}:PersonalisedName>"
             )
 
+        attrib = attrib.replace("'", '"')
+        data = data.replace("'", '"')
         outfile += f"{indent}<{label}{attrib}>"
+        outfile += intern
         outfile += data
         outfile += f"{indent}</{label}>"
 
     if isinstance(item_or_scheme, VtlScheme):
-        outfile += f" vtlVersion={item_or_scheme.vtl_version!r}"
-
-    outfile = outfile.replace("'", '"')
+        outfile += f' vtlVersion="{item_or_scheme.vtl_version}"'
 
     return outfile
 
--- a/src/pysdmx/io/xml/__structure_aux_writer.py
+++ b/src/pysdmx/io/xml/__structure_aux_writer.py
@@ -210,8 +210,8 @@ ANNOTATION_WRITER = OrderedDict(
     {
         "title": "AnnotationTitle",
         "type": "AnnotationType",
-        "text": "AnnotationText",
         "url": "AnnotationURL",
+        "text": "AnnotationText",
     }
 )
 
@@ -326,8 +326,8 @@ def __write_annotable(
         if annotation.id is None:
             outfile += f"{child2}<{ABBR_COM}:Annotation>"
         else:
-            outfile += f"{child2}<{ABBR_COM}:Annotation id={annotation.id!r}>"
-        outfile = outfile.replace("'", '"')
+            annotation_id = __escape_xml(annotation.id)
+            outfile += f'{child2}<{ABBR_COM}:Annotation id="{annotation_id}">'
 
         for attr, label in ANNOTATION_WRITER.items():
             if getattr(annotation, attr, None) is not None:
--- a/tests/io/xml/sdmx21/writer/test_structures_writing.py
+++ b/tests/io/xml/sdmx21/writer/test_structures_writing.py
@@ -2230,6 +2230,105 @@ def test_metadata_structure_21_raises(complete_header):
         write([msd], header=complete_header, prettyprint=True)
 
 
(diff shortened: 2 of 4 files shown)


javihern98:
@albertohernandez1995 Please provide evidence as well this does not happen in Name and Description

albertohernandez1995:
@javihern98 Checked this, and you were right.

For everything non-VTL it's fine: I wrote codelists, concept schemes, agency/provider/consumer schemes, category schemes, DSDs, dataflows, hierarchies and constraints with apostrophes in both `name` and `description`, at scheme and item level, in 2.1/3.0/3.1. All clean, all round-trip unchanged. That's structural — `__write_nameable` already quotes these explicitly, and in those writers the replacement only ever touches the `Attributes` string, not the buffer holding Name/Description.

The exception is VTL scheme **items**:

```xml
<com:Name xml:lang="en">SCHEME'NAME</com:Name>   <!-- fine -->
<com:Name xml:lang="en">ITEM"NAME</com:Name>     <!-- corrupted -->
```

`_write_vtl` puts the item's Name/Description into `data`, appends it to `outfile`, then does `outfile = outfile.replace("'", '"')` over the whole buffer. Same bug, different function. Hits all six VTL item types in all three versions.

It's pre-existing — I reproduced it on `develop` with this PR's changes removed. The fix is the same shape: scope that replacement to `attrib` rather than `outfile`.

I'd rather do it as a separate issue since it's another function and needs its own tests (and it interacts with the `&apos;` escaping from #615). Happy to fold it in here instead if you prefer.

javihern98:
Yes, @albertohernandez1995 please include it here

albertohernandez1995:
@javihern98 Done, included here as you asked.

javihern98 (commented):

sosna (approved):
This looks good, thanks, @albertohernandez1995

PR #684, opened 2026-08-31
Strict RFC 3339 date-time validation and UTC-normalized SDMX datetimes

### LINKED ISSUE #639: SDMX-JSON: strict date-time validation + normalize SDMX datetimes to timezone-aware

javihern98:
Closes #639.

## Changes

- Added `rfc3339-validator` to the `json` and `all` extras, enabling jsonschema's strict RFC 3339 `date-time` format check. The `all` extra also gains `jsonschema`, which it was missing even though JSON validation is enabled by default when reading SDMX-JSON messages.
- SDMX datetimes are normalized to timezone-aware UTC (timezone-less datetimes are assumed to be in UTC): facet `start_time`/`end_time`, value-map validity and `meta.prepared`, across the SDMX-JSON reader/writer, the Fusion-JSON reader and the SDMX-ML reader/writer. The SDMX-JSON value-map writer no longer drops the timezone, so writer output is RFC 3339 compliant and round trips preserve model equality.
- Schema failures caused by the strict `date-time` check are now reported as a `UserWarning` pointing at the offending values, instead of raising `Invalid` (SDMX itself allows timezone-less datetimes); any other failure still raises `Invalid`. Documented in the IO tutorial.
- A date-only 2.1 `meta.prepared` is no longer falsely rejected (it previously matched both the `date` and the unchecked `date-time` branches of the `oneOf`).

Note: artefact-level validity read from SDMX-ML or item schemes (e.g. `ConceptScheme.valid_from`) intentionally stays as-is; timezone-less values written to SDMX-JSON are surfaced by the new warning and could be normalized in a follow-up.

## Verification

```python
import warnings
from pysdmx.io.writer import write_sdmx

MSG = """{
  "meta": {"id": "T", "schema": "https://raw.githubusercontent.com/sdmx-twg/sdmx-json/develop/structure-message/tools/schemas/2.0.0/sdmx-json-structure-schema.json",
           "prepared": "2025-01-01T00:00:00Z", "sender": {"id": "ZZZ"}},
  "data": {"conceptSchemes": [{
    "id": "CS", "name": "CS", "version": "1.0", "agencyID": "TEST",
    "concepts": [{"id": "DIM3", "name": "Concept",
      "coreRepresentation": {"format": {"dataType": "DateTime", "startTime": "2000-01-01T00:00:00"}}}]}]}
}"""

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    msg = read_sdmx(MSG)
print(w[0].message)
# UserWarning: ... $.data.conceptSchemes.0.concepts.0.coreRepresentation.format.startTime:
# '2000-01-01T00:00:00' is not an RFC 3339 date-time ...
print(msg.structures[0].items[0].facets.start_time)
# 2000-01-01 00:00:00+00:00  (normalized to UTC)

out = write_sdmx(msg.structures, Format.STRUCTURE_SDMX_JSON_2_0_0, header=msg.header)
with warnings.catch_warnings():
    warnings.simplefilter("error")
    read_sdmx(out)  # writer output is RFC 3339 compliant: no warning, no error
print("round trip validates cleanly")
```


src/pysdmx/util/__init__.py
```diff
@@ -19,6 +20,35 @@
 semver_final_pattern = re.compile(rf"^[1-9]\d*\.{_SEMVER_NUM}\.{_SEMVER_NUM}$")
 
 
+@overload
+def to_utc(value: datetime) -> datetime: ...
```
sosna, line 24: Why do we need these 2 @overload functions?

I would have expected the function on line 31 to be sufficient?

docs/howto/io_structure.rst
```diff
@@ -18,6 +18,13 @@ In this tutorial, we learn how to read SDMX Structures messages using the
     For SDMX-JSON structural validation you need to install the `pysdmx[json]` extra.
     ``validate=True`` is the default behaviour on read_sdmx and get_datasets.
 
+    SDMX-JSON validation strictly checks RFC 3339 ``date-time`` values,
+    which require timezone information. As SDMX itself allows datetimes
+    without timezone information, such datetimes are reported with a
+    ``UserWarning`` instead of an error, and the message is read anyway.
+    ``pysdmx`` normalizes SDMX datetimes to timezone-aware UTC datetimes
```
sosna, line 25: > ``pysdmx`` normalizes SDMX datetimes to timezone-aware UTC datetimes

Is this comment correct? I don't think pysdmx does this with all aware datetimes. For example:

```python
from datetime import datetime, timedelta, timezone

from pysdmx.model import Organisation
from pysdmx.model.message import Header

prepared = datetime(
    2026, 1, 1, 12, tzinfo=timezone(timedelta(hours=2))
)
header = Header(
    id="TEST",
    prepared=prepared,
    sender=Organisation(id="TEST"),
)

assert header.prepared.isoformat() == "2026-01-01T12:00:00+02:00"
```

Much more importantly, **I don't think we should change aware datetimes**. If a user supplies an aware datetime in a timezone other than UTC, this is fine, i.e. there is no ambiguity and there is no need to force a conversion to UTC.

However, in case of naïve datetimes then, yes, we default to UTC.

What do you think?

src/pysdmx/util/__init__.py
```diff
@@ -19,6 +20,35 @@
...
+        value: The datetime to be normalized (or None).
+
+    Returns:
+        A timezone-aware datetime in UTC, or None if the input is None.
+    """
+    if value is None:
+        return None
+    if value.tzinfo is None:
+        return value.replace(tzinfo=timezone.utc)
+    return value.astimezone(timezone.utc)
```
sosna, line 49: As mentioned in my previous comment, I'm not sure why we should change the time zone received in input. I think it's OK to default to UTC when receiving naïve datetimes. But, at the same time, if we have received an aware datetime, I'm not sure why we should forcibly convert it to UTC, unless I'm overlooking something...

sosna (changes requested):
Thanks, @javihern98. I think we should probably discuss on this one, unless you agree with my main comment.

src/pysdmx/util/__init__.py
javihern98: I replaced the overload with a bound TypeVar and added a comment. AI tests shows mypy is always failing and several type ignore statements are needed if we keep only the base function.

docs/howto/io_structure.rst
javihern98: Yes, I agree, it is best to keep the timezone information if previded by the user, elsewise make it UTC. I changed the function from `to_utc` to `ensure_tz_aware`, this code snippet now works as expected

src/pysdmx/util/__init__.py
javihern98: Agree, I changed the function

javihern98:
I am now ensuring parity regarding the timezone awareness in both SDMX-ML and SDMX-JSON. The PR needs to be fully reviewed again due to the many files changes (most of them adding just the timezone)

src/pysdmx/io/json/sdmxjson2/messages/code.py
```diff
@@ -24,7 +23,7 @@
     HierarchyAssociation,
     LevelType,
 )
-from pysdmx.util import find_by_urn, is_final, parse_item_urn
+from pysdmx.util import ensure_tz_aware, find_by_urn, is_final, parse_item_urn
 
 _VAL_FMT = "%Y-%m-%dT%H:%M:%S%z"
```
sosna: Maybe this should be defined in one place? It has already been defined in the FusionJSON message. Maybe we could define constants like _VAL_TS_FMT and _VAL_NAIVE_TS_FMT, in pysdmx.util?

src/pysdmx/util/__init__.py
```diff
@@ -19,6 +20,31 @@
...
+
+    Args:
+        value: The datetime to be checked (or None).
+
+    Returns:
+        A timezone-aware datetime, or None if the input is None.
+    """
+    if value is None:
+        return value
+    if value.tzinfo is None:
```
sosna: Apparently, this is marginally safer:

```python
if value.tzinfo is None or value.utcoffset() is None:
        return value.replace(tzinfo=timezone.utc)
```

In case someone creates a datetime with a custom tzinfo class not providing a utc offset... But maybe it's a bit far fetched :)

sosna (dismissed):
Thanks, @javihern98, this looks very good. I have two minor comments, but, I have already approved the PR so feel free to ignore if you disagree :)

src/pysdmx/util/__init__.py
javihern98: I agree, replaced and added a dedicated test

src/pysdmx/io/json/sdmxjson2/messages/code.py
javihern98: Yes, I moved it to pysdmx.util and simplified the code

sosna (approved):
Looks good! Thanks a lot, @javihern98!

PR #687, opened 2026-09-01
Fix crashes on empty and omitted structure containers when reading SDMX-ML

### LINKED ISSUE #685: read_sdmx raises TypeError on an empty structure catalog (SDMX-ML 3.0)
### Environment
- pysdmx 1.19.0 (installed as `pysdmx[xml]`)
- Python 3.12, Windows

### Summary
`read_sdmx` on a well-formed SDMX-ML 3.0 Structure message whose artefact
container element is empty (e.g. `<str:Dataflows/>`) raises
`TypeError: 'NoneType' object is not subscriptable` instead of returning a
`Message` with no artefacts of that type.

This is a real response shape: the Eurostat SDMX 3.0 endpoint answers
`GET .../structure/dataflow/ESTAT` with exactly this when there is no bulk
dataflow catalogue.

### Reproduction
```python
from io import BytesIO
import pysdmx.io as sio

EMPTY = b"""<?xml version="1.0" encoding="UTF-8"?>
<m:Structure xmlns:m="http://www.sdmx.org/resources/sdmxml/schemas/v3_0/message"
 xmlns:s="http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure">
  <m:Header><m:ID>ID</m:ID><m:Test>true</m:Test>
    <m:Prepared>2026-01-01T00:00:00Z</m:Prepared><m:Sender id="ESTAT"/></m:Header>
  <m:Structures><s:Dataflows/></m:Structures>
</m:Structure>"""

sio.read_sdmx(BytesIO(EMPTY), validate=False)
```

### Actual
```
File ".../pysdmx/io/xml/__structure_aux_reader.py", line 2257, in __format_schema
    json_element[item] = add_list(json_element[item])
                                  ~~~~~~~~~~~~^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

### Expected
A `Message` with `msg.get_dataflows() == ()` (or a typed pysdmx error).

### Cause / suggested fix
`xmltodict` maps the empty `<str:Dataflows/>` element to `None`, so
`format_structures.process_structure` calls `formatter(None)` and
`__format_schema` then indexes into `None`. A guard in `process_structure`
(~line 2350) would cover every artefact type:

```python
if key in json_meta and json_meta[key] is not None:
    formatted = formatter(json_meta[key])
```

Happy to open a PR if you'd like.

albertohernandez1995:
## Summary

Closes #685. `read_sdmx` crashed with a bare `TypeError` on an SDMX-ML Structure message whose artefact container is empty. Reproduced verbatim from the issue before changing anything, then traced to `StructureParser`: `xmltodict` maps an empty XML element to `None`, and that `None` reached the structure formatters unguarded.

Investigating it turned up a **second, distinct crash** on a related shape that the issue does not mention, and which matters more because that shape is schema-valid:

| Input | Schema-valid? | Before | After |
| ----- | ------------- | ------ | ----- |
| `<m:Structures><s:Dataflows/></m:Structures>` | **No** — the XSD requires at least one child | `TypeError: 'NoneType' object is not subscriptable` | reader returns empty; `read_sdmx` returns an empty `Message` |
| `<m:Structures/>` — every container omitted | **Yes** — this is how a service reports an empty catalogue | `AttributeError: 'NoneType' object has no attribute 'get'` | reader returns empty; `read_sdmx` returns an empty `Message` |
| `<m:Structures>` omitted altogether — header only | **Yes** — `Structures` is `minOccurs="0"` in 2.1, 3.0 and 3.1 | `KeyError: 'Structures'` | reader returns empty; `read_sdmx` returns an empty `Message` |

The first two are fixed by one guard each, placed at the single choke point every artefact type and every SDMX-ML version passes through; the third (found during review) by letting the three version readers hand `None` to that same choke point. Following the review discussion, `read_sdmx` then returns the resulting empty `Message` instead of rejecting it, and the SDMX-JSON structure reader accepts a meta-only message so the same holds there.

## What's included

**`io/xml/__structure_aux_reader.py`** — two guards in `StructureParser`:

- **`process_structure`** — `if json_meta.get(key) is not None:` in place of `if key in json_meta:`, so a container that is present but empty is treated exactly like an absent one instead of handing `None` to a formatter that indexes into it.
- **`format_structures`** — accepts `Optional[Dict[str, Any]]` and normalises `None` to `{}`, covering an empty `<mes:Structures/>`. All three version readers call this with `dict_info[STRUCTURE][STRUCTURES]`, so one guard fixes 2.1, 3.0 and 3.1.

**`io/xml/sdmx{21,30,31}/reader/structure.py`** — `dict_info[STRUCTURE].get(STRUCTURES)` instead of indexing, so a message that omits `<mes:Structures>` altogether reaches `format_structures` as `None` and is covered by the guard above instead of raising a bare `KeyError`.

**`io/json/sdmxjson2/messages/structure.py`** — `data` gets `field(default_factory=JsonStructures)`. The SDMX-JSON structure schema does not require `data` (2.0.0 has no top-level `required`; 2.1 requires only `meta`), yet a meta-only message was rejected with a misleading `Invalid("...could not be read as SDMX-JSON 2.0.0 or 2.1.0 structure message")`. It now decodes to an empty `Message`. Writer output is unchanged: msgspec does not omit `default_factory` fields, so an empty message is still written with `"data": {}`.

**`io/reader.py`** — `read_sdmx` no longer applies the `Empty SDMX Message` guard to structure messages (SDMX-ML 2.1/3.0/3.1 and SDMX-JSON 2.0/2.1). A structure message without artefacts is returned as a `Message` carrying its header and an empty `structures` sequence, so `msg.get_dataflows() == []` as the issue asked for. Data, submission and reference-metadata messages keep raising `Invalid("Empty SDMX Message")`. The `Raises` section now lists what is actually raised: `Invalid` for content-less data, submission and SDMX-ML reference-metadata messages, `NotFound` for an SDMX-JSON reference-metadata message without reports (raised by `get_reports()` before the guard), and `NotImplemented` for detected-but-unsupported flavours.

Supporting changes:

- `tests/io/xml/sdmx{21,30,31}/reader/test_reader.py` — three tests per version. `test_read_empty_structure_containers_*` covers containers that are present but empty (`Dataflows`, `Codelists`, `ConceptSchemes` and `AgencySchemes`/`OrganisationSchemes`, the last exercising the `not iterable` variant through `__format_orgs`), read with `validate=False`; `test_read_no_structure_containers_*` covers the `<m:Structures/>` shape with `validate=True`; `test_read_header_only_structure_message_*` (new sample `structures_header_only.xml`) covers the message with no `<mes:Structures>` at all, also with `validate=True`. Each asserts both layers: `read_structure` returns an empty sequence, and `read_sdmx` returns a `Message` with a header, `structures == []` and `get_dataflows() == []`.
- `tests/io/test_general_reader.py` — `test_empty_result` now asserts the empty `Message` instead of `Invalid`. New `test_get_json20_empty_structure` (with `samples/empty_structure_message.json`, a structure message whose `data` object is empty) pins the same behaviour for the SDMX-JSON structure path, which goes through the same guard; `test_get_json20_structure_meta_only` (with `samples/structure_message_meta_only.json`) covers the meta-only message. New `test_empty_data_message` (with `samples/empty_data_message.xml`, a schema-valid `GenericData` message without any `DataSet`) pins that data messages still raise `Invalid("Empty SDMX Message")`, which also keeps that branch covered.

## Design decisions for review.

- **The scope is much wider than reported, and is fixed in full.** The issue describes `<str:Dataflows/>` in SDMX-ML 3.0; in fact every artefact container fails in all three versions, in two error variants (`not subscriptable`, and `not iterable` for `AgencySchemes`/`OrganisationSchemes`). All share one root cause, so the single guard covers them. Confirm you want the whole surface fixed here rather than narrowed to the reported case.
- **An empty structure message returns an empty `Message` rather than raising.** Agreed in the review thread: an empty catalogue is a valid server response, while `Invalid` is a non-retriable client error, and users bulk-loading structure files or going through the FMR client are better served by an empty `Message`. The guard is kept for data, submission and reference-metadata messages, where an empty payload has no comparable "empty catalogue" meaning and `get_datasets` already reports `No data found in the data message`; extending the same treatment there is tracked separately in #694.
- **Two premises in the issue are not quite right, and neither changes the fix.** The document is well-formed XML but **not** schema-valid — the container is `minOccurs="0"`, so a service with no dataflows should omit `<str:Dataflows>` entirely; emitting it empty looks like a server-side bug at Eurostat. Consequently the reported crash was only reachable with `validate=False`; with `validate=True` pysdmx already raised a precise validation error. The shape Eurostat *should* be sending crashed harder, in both validate modes.

## How it was tested

Reproduced first: the issue's snippet was run verbatim on `develop` and raised the reported `TypeError` at the reported line, via `process_structure`. A sweep over 9 container types × 3 SDMX-ML versions then established the true blast radius — 27/27 raising.

Full suite: 5144 passed, 100% line and branch coverage; ruff format, ruff check and mypy clean.

Every new or changed test was watched failing before its fix and passing after it: the container-guard tests against `develop`, and the empty-`Message` assertions against the previous `Invalid` guard (7 failures, then 8 passes once `read_sdmx` was changed; the SDMX-JSON tests were likewise run against the pre-change code first, and the header-only SDMX-ML tests failed with the bare `KeyError` before the reader change), so they pin the behaviour rather than merely passing.

Schema validity of both shapes was checked directly against the bundled `sdmxschemas` XSDs for v2_1, v3_0 and v3_1, which is what establishes that the omitted-container case is the valid one and the empty-container case is not.


javihern98 (commented):

tests/io/xml/sdmx21/reader/test_reader.py
```diff
@@ -1618,3 +1618,26 @@ def test_category_scheme_21_enrichment_edge_cases(samples_folder):
...
+        read_sdmx(input_str, validate=False)
+
+
+def test_read_no_structure_containers_21(samples_folder):
+    data_path = samples_folder / "structures_no_containers.xml"
+    input_str, _ = process_string_to_read(data_path)
+
+    assert len(read_structure(input_str, validate=True)) == 0
+
+    with pytest.raises(Invalid, match="Empty SDMX Message"):
```
sosna: If I understood the description correctly, in this case, the input file is valid. If this is correct, then, maybe a NotFound error would be more appropriate, in the sense that, while the file is OK, we could not find any structure in it?

tests/io/xml/sdmx30/reader/test_reader.py
```diff
@@ -1506,3 +1506,26 @@ def test_categorisation_30(samples_folder):
...
+        read_sdmx(input_str, validate=False)
+
+
+def test_read_no_structure_containers_30(samples_folder):
+    data_path = samples_folder / "structures_no_containers.xml"
+    input_str, _ = process_string_to_read(data_path)
+
+    assert len(read_structure(input_str, validate=True)) == 0
+
+    with pytest.raises(Invalid, match="Empty SDMX Message"):
```
sosna: Same comment re. Invalid vs. NotFound

tests/io/xml/sdmx31/reader/test_reader.py
```diff
@@ -475,3 +475,26 @@ def test_categorisation_31(samples_folder):
...
+        read_sdmx(input_str, validate=False)
+
+
+def test_read_no_structure_containers_31(samples_folder):
+    data_path = samples_folder / "structures_no_containers.xml"
+    input_str, _ = process_string_to_read(data_path)
+
+    assert len(read_structure(input_str, validate=True)) == 0
+
+    with pytest.raises(Invalid, match="Empty SDMX Message"):
```
sosna: Same comment re. Invalid vs. NotFound

sosna (commented):
It looks good, @albertohernandez1995, thanks. I just have a question re. the use of NotFound vs. Invalid.

tests/io/xml/sdmx21/reader/test_reader.py
javihern98: On a second look, is this really an error? @sosna @stratosn Given that the XSD allow for completely empty SDMX messages and now that we support most of the objects...

sosna: What do you suggest? Returning a structure message with no artefacts? I guess that's OK too. I'm not sure it's better than NotFound, but I definitely prefer it to Invalid :).

javihern98: Yes, I believe it would be better, mainly for the users who make bulk loading of many structures files or using the FMR client, I believe it would be best to just return the empty message.

Shall we proceed?

sosna: That sounds good to me, thanks @javihern98 .

javihern98: Ok, limited the scope for this issue for Structure messages, a follow up issue will look into the other messages. I will update the PR description

sosna (approved):
This looks good. Thanks, @albertohernandez1995

PR #688, opened 2026-09-01
Keep the component enumeration reference when the codelist is not in the message

### LINKED ISSUE #686: Component enumeration reference is dropped when the codelist is not in the message
### Environment
- pysdmx 1.19.0 (installed as `pysdmx[xml]`)
- Python 3.12, Windows

### Summary
For a coded DSD component whose `<str:LocalRepresentation><str:Enumeration>`
holds a codelist URN, when that codelist is **not** present in the same
message, pysdmx sets **both** `component.enumeration` and `component.enum_ref`
to `None`. The reference is lost entirely: a consumer cannot tell the
component is coded, nor which codelist to fetch to resolve it.

This affects every coded component of a `detail=full` DSD response (the common
case — referenced codelists are a separate request).

When the codelist **is** in the message, `component.enumeration` is correctly
the `Codelist`.

### Reproduction
```python
from io import BytesIO
import pysdmx.io as sio

DSD = """<?xml version="1.0" encoding="UTF-8"?>
<m:Structure xmlns:m="http://www.sdmx.org/resources/sdmxml/schemas/v3_0/message"
 xmlns:s="http://www.sdmx.org/resources/sdmxml/schemas/v3_0/structure"
 xmlns:c="http://www.sdmx.org/resources/sdmxml/schemas/v3_0/common">
  <m:Header><m:ID>ID</m:ID><m:Test>true</m:Test>
    <m:Prepared>2026-01-01T00:00:00</m:Prepared><m:Sender id="ZZZ"/></m:Header>
  <m:Structures>
    <s:DataStructures>
      <s:DataStructure agencyID="ZZZ" id="DSD_TEST" version="1.0">
        <c:Name xml:lang="en">Test</c:Name>
        <s:DataStructureComponents>
          <s:DimensionList id="DimensionDescriptor">
            <s:Dimension id="FREQ" position="1">
              <s:ConceptIdentity><Ref agencyID="ZZZ" maintainableParentID="CS"
                maintainableParentVersion="1.0" id="FREQ" class="Concept"
                package="conceptscheme"/></s:ConceptIdentity>
              <s:LocalRepresentation>
                <s:Enumeration>urn:sdmx:org.sdmx.infomodel.codelist.Codelist=ZZZ:CL_FREQ(1.0)</s:Enumeration>
              </s:LocalRepresentation>
            </s:Dimension>
          </s:DimensionList>
          <s:MeasureList id="MeasureDescriptor">
            <s:Measure id="OBS_VALUE"><s:ConceptIdentity><Ref agencyID="ZZZ"
              maintainableParentID="CS" maintainableParentVersion="1.0"
              id="OBS_VALUE" class="Concept" package="conceptscheme"/></s:ConceptIdentity></s:Measure>
          </s:MeasureList>
        </s:DataStructureComponents>
      </s:DataStructure>
    </s:DataStructures>
  </m:Structures>
</m:Structure>""".encode()

msg = sio.read_sdmx(BytesIO(DSD), validate=False)
freq = next(c for c in msg.get_data_structure_definitions()[0].components
            if c.id == "FREQ")
print(repr(freq.enum_ref), repr(freq.enumeration))   # None None
```

### Actual
`freq.enum_ref is None` and `freq.enumeration is None` — the
`ZZZ:CL_FREQ(1.0)` reference is gone.

### Expected
`freq.enum_ref` retains the URN, as it already does for `MetadataAttribute`s.

### Cause / suggested fix
`__format_component` (`src/pysdmx/io/xml/__structure_aux_reader.py`, ~line 1010)
calls `__format_local_rep` but, unlike `__format_metadata_attribute`
(~lines 1226–1230), never extracts `local_enum_ref`. Mirroring that block:

```python
if LOCAL_REP in comp:
    local_enum_ref = self.__local_enum_ref(comp[LOCAL_REP])
    if local_enum_ref is not None:
        comp["local_enum_ref"] = local_enum_ref
    self.__format_local_rep(comp)
```

`Component.local_enum_ref` / the `Component.enum_ref` property and the SDMX-ML
writer path already exist. (This covers the SDMX-ML 3.0 URN-text form; the
2.1 `<Ref>` form handled by `__local_enum_ref` would need the same treatment.)

Happy to open a PR.

albertohernandez1995:
## Summary

Closes #686. For a coded DSD component whose `<str:LocalRepresentation><str:Enumeration>` holds a codelist reference, pysdmx set **both** `component.enumeration` and `component.enum_ref` to `None` whenever that codelist was not in the same message. The reference was lost outright, so a consumer could not tell the component was coded nor which codelist to fetch — and this is the normal shape of a `detail=full` DSD response, since referenced codelists are a separate request.

Reproduced verbatim from the issue before changing anything. The model side (`Component.local_enum_ref` and the `Component.enum_ref` property) was already in place, and the **SDMX-JSON 2.0 reader already does this correctly**, so this was an XML-path gap rather than a missing feature.

## What's included

**`io/xml/__structure_aux_reader.py`**

- **`__format_local_rep`** — captures the enumeration URN *before* `__format_representation` consumes the local representation, and stores it as `local_enum_ref`. Placed here rather than in the caller so that DSD components and MSD attributes share one path.
- **`__format_metadata_attribute`** — its duplicate of that logic is removed; it now just picks `local_enum_ref` up alongside `local_codes` / `local_dtype` / `local_facets`.
- **`__local_enum_ref`** — generalised. It previously assumed the SDMX-ML 3.x URN-text form and returned `str(local_rep[ENUM])`, which for the other shapes would have stringified a dict. It now returns a full URN from all three real forms: 3.x URN text, a 2.1 `<Ref>` element, and a `<URN>` element (the last two both already handled by `__format_representation`).

**`io/xml/__structure_aux_writer.py`**

- **`__write_representation`** — falls back to `local_enum_ref` when `local_codes` is `None`, and the `EnumerationFormat` / `TextFormat` decision now keys off whether *any* enumeration was written.
- **`__write_enumeration_ref`** — gains a `references_30` parameter so it emits URN text for 3.x and a `<Ref>` element for 2.1, mirroring `__write_enumeration`. Its previous MSD-only callers are unaffected by the default.

**`io/xml/__tokens.py`** — adds the `LOCAL_ENUM_REF` token next to its siblings.

Supporting changes:

- `tests/io/xml/sdmx{21,30,31}/reader/test_reader.py` — 8 tests: the reference is retained with the codelist absent (all three versions); a full read → write → read round-trip that validates against the XSD (all three); the 2.1 `<URN>` reference form; and a no-regression test that with the codelist in the message `enumeration` is still the `Codelist` object.

## Design decisions for review.

- **The writer fix is included, and the reader fix alone would not have been enough.** `__write_representation` only ever emitted an enumeration from `local_codes`, so once the model carried the reference a read/write cycle still discarded it — and wrote `<str:TextFormat/>` where the enumeration belonged, turning a coded component into an uncoded one. Fixing only the reader would have moved the data loss one step later rather than removing it. Confirm you want both here, or I will split the writer into its own issue.
- **The two forms the issue flags as needing "the same treatment" are both covered.** `__local_enum_ref` now handles the 2.1 `<Ref>` element and the `<URN>` element as well as 3.x URN text, all returning a full URN so `enum_ref` is version-independent and matches the SDMX-JSON 2.0 reader.
- **The extraction sits in `__format_local_rep`, not `__format_component` as the issue suggests.** The suggested placement takes `__format_component` to a cyclomatic complexity of 12 against the project limit of 10. The chosen placement stays within the limit and removes the existing duplication in `__format_metadata_attribute`. Confirm this is preferred over a `noqa`.
- **`enum_ref` is now also populated when the codelist *is* in the message,** where it previously fell through to the concept's `enum_ref` and was usually `None`. This makes `enum_ref` reliable in both cases, but it is a visible behaviour change for callers that used `enum_ref is None` to infer "the codelist is inlined" — `enumeration` remains the correct signal for that. Confirm whether it warrants a changelog note for 1.20.0.

## How it was tested

Reproduced first: the issue's snippet was run verbatim on `develop` and returned `None None` as reported.

Each of the 8 new tests was confirmed to fail with both fixes reverted and to pass with them restored.

Round-trip output was checked against the bundled `sdmxschemas` XSDs for v2_1, v3_0 and v3_1 — all three re-read with `validate=True` and the reference intact, which is what establishes that the newly emitted `<str:Enumeration>` is well-formed in each version's own reference style.

Full suite: 5122 passed, 100% branch coverage; ruff format, ruff check and mypy clean. Reaching 100% required covering the `<URN>` reference branch, which no existing test exercised — that gap is what surfaced the third reference form.


javihern98 (commented):

sosna (approved):
This looks good, thanks, @albertohernandez1995!

PR #695, opened 2026-09-14
Read valid empty data and reference metadata messages as empty Messages

### LINKED ISSUE #694: read_sdmx still rejects valid empty data and reference metadata messages
Follow-up to #685 / #687, which made `read_sdmx` return an empty `Message` for structure messages without artefacts. The same reasoning (an empty payload is valid SDMX, and `Invalid` is a non-retriable client error) applies to the other message types, which were deliberately left out of #687 to keep its scope to structures.

### Current behaviour

| Empty message | Schema-valid? | `read_sdmx` today | Raised by |
| --- | --- | --- | --- |
| SDMX-ML data with no `DataSet` (2.1 Generic and StructureSpecific, 3.0, 3.1) | Yes, `DataSet` is `minOccurs="0"` | `Invalid("Empty SDMX Message")` | guard in `io/reader.py` |
| SDMX-ML 3.0/3.1 GenericMetadata with no `MetadataSet` | Yes, `minOccurs="0"` | `Invalid("Empty SDMX Message")` | guard in `io/reader.py` (the reader itself already returns `[]`) |
| SDMX-JSON 2.0/2.1 metadata with `"metadataSets": []` | Yes | `NotFound("No metadata reports were found in the message.")` | `MetadataMessage.get_reports()`, called by `read_sdmx` |
| SDMX-JSON metadata with `data` or `metadataSets` omitted | Yes, the schemas require only `meta` | misleading `Invalid("...could not be read as ... reference metadata message")` | msgspec: `JsonMetadataMessage.data` and `JsonMetadataSets.metadataSets` have no defaults |
| SDMX-ML 2.1 RegistryInterface with an empty `SubmitStructureResponse` | No, `SubmissionResult` is required | bare `TypeError` when `validate=False` | `io/xml/sdmx21/reader/submission.py` |

All "yes" rows were checked with `validate=True` against the bundled XSDs / JSON schemas.

### Proposal

- Return an empty `Message` (`data=[]`, `reports=[]`) instead of raising, mirroring #687. `get_datasets()` keeps its own `No data found in the data message` check.
- In the SDMX-JSON metadata branch of `read_sdmx`, read `ref_msg.reports or []` instead of calling `get_reports()`.
- Give `JsonMetadataMessage.data` and `JsonMetadataSets.metadataSets` `default_factory` defaults, as done for `JsonStructureMessage.data` in #687. Encoding is unchanged, since msgspec does not omit `default_factory` fields.
- Decide whether `Message.get_reports()` should return `[]` for an empty list, as `get_datasets()` and the structure getters do; today it raises `NotFound`.
- Registry: raise a typed `Invalid` for the schema-invalid empty response instead of a `TypeError`.

javihern98:
Closes #694. Follow-up to #687, which made `read_sdmx` return an empty `Message` for structure messages without artefacts. The same treatment now applies to the other message types.

### Changes

- **`read_sdmx`** no longer raises `Invalid("Empty SDMX Message")`. A data message without any `DataSet` (SDMX-ML 2.1 Generic and StructureSpecific, 3.0, 3.1) returns a `Message` with `data == []`, and a reference metadata message without reports (SDMX-ML 3.0/3.1 GenericMetadata, SDMX-JSON 2.0/2.1) returns one with `reports == []`. `get_datasets()` keeps its own `No data found in the data message` check. SDMX-CSV is unaffected, as its readers already reject header-only input.
- **SDMX-JSON metadata models**: `JsonMetadataMessage.data` and `JsonMetadataSets.metadataSets` get `default_factory` defaults (only `meta` is required by the schemas), so a message with `data` or `metadataSets` omitted no longer fails with a misleading `Invalid`. Encoding is unchanged, since msgspec never omits `default_factory` fields.
- **`Message.get_reports()` and `MetadataMessage.get_reports()`** return an empty sequence for a message with `reports == []` and raise `NotFound` only when `reports is None`, matching `get_datasets()` and the structure getters.
- **SDMX-ML 2.1 registry reader**: a `SubmitStructureResponse` without `SubmissionResult` raises a typed `Invalid` instead of a bare `TypeError` when `validate=False` (the XSD already rejects it when `validate=True`).


sosna (approved):
This looks good. Thanks, @javihern98

